PR #11, opened 2019-03-14
Update README.md

lslachciak:
[Issue] N/A
[Problem] No description about releases
[Solution] Add separate site in docs about releases

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -1,0 +1,18 @@
+# TAU releases
+
+## Release download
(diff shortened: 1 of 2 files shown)


docs/releases.md
```diff
@@ -0,0 +1,18 @@
+# Tau releases
```
TomaszLukawskiSam: Tau -> TAU

lslachciak: Done

PR #45, opened 2019-04-19
Listview (waearble): js errors when building ArcListview

TomaszLukawskiSam:
[Solution]
To solve this issue has needed changes in engine.
Added new option to define widget.
Option: buildOptions.requireMatchSelector means that widget can build
 only on defined selector.


wszczepanski97 (approved):
LGTM :+1:

PR #55, opened 2019-04-29
Fix listview divider style on non built list

pkaczmarczy:
[Problem] List divider in in non built list was treated like group index
[Solution] Use more restricted selector in less file


pkaczmarczy:
Done

pkaczmarczy:
This commit can be tested in wearable UIComponents app.
To disable list widgets building remove 32nd line ("<script src="js/circle-helper.js"></script>") from:
examples/wearable/UIComponents/index.html and restart app

Issue #58, opened 2019-05-06
GalleryUI (example from SDK): content is cut on bottom


TomaszLukawskiSam:
![image](https://user-images.githubusercontent.com/29534410/57231650-38bf4d80-701b-11e9-802e-411b826963be.png)

TomaszLukawskiSam:
Issue has been already solved.

PR #59, opened 2019-05-06
Page (core): change calculating of content size

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/58 #58 
[Problem] GalleryUI (example from SDK): content is cut on bottom
[Solution] Method for calculating size of content has been changed.
If footer is not displayed then footer has size zero.

### DIFF
--- a/src/js/core/widget/core/Page.js
+++ b/src/js/core/widget/core/Page.js
@@ -456,7 +456,7 @@
 					}
 
 					if (footer) {
-						bottom = utilsDOM.getElementHeight(footer);
+						bottom = footer.getBoundingClientRect().height;
 						contentStyle.marginBottom = bottom + "px";
 						contentStyle.paddingBottom = (-bottom) + "px";
 					}


src/js/core/widget/core/Page.js
```diff
@@ -456,9 +457,12 @@
 					}
 
 					if (footer) {
-						bottom = utilsDOM.getElementHeight(footer);
-						contentStyle.marginBottom = bottom + "px";
-						contentStyle.paddingBottom = (-bottom) + "px";
+						footerRect = footer.getBoundingClientRect();
+						if (footerRect) {
```
pkaczmarczy: According to w3c (https://www.w3.org/TR/2009/WD-cssom-view-20090804/#the-getclientrects-and-getboundingclient) this check is unnecessary - the getBoundingClientRect method always returns an ClientRect object

TomaszLukawskiSam: Fixed.

Issue #69, opened 2019-05-08
Header and Footer have too strong selectors to hide by developers


TomaszLukawskiSam:
Example MiniGallery (SDK)
Header in image preview page has display class that is managed by the app. It hides the header in preview. The css selector is too weak and it cannot overcome TAU visibility settings. The app should use different mechanism to hide header, or find way to make selector stronger.

TomaszLukawskiSam:
For resolve the issue the app has to be changed.
file: MiniGallery/css/style.css

- add style
```
.ui-page .ui-header.hidden {
    display: none;
}
```

PR #70, opened 2019-05-08
Header/Footer too strong selectors to hide elements

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/69
[Problem] Header and Footer have too strong selectors to hide by developers
[Solution] This solution is not good but it seems only way to
fix apps where developer want to hide header or footer by class ".hidden".


TomaszLukawskiSam:
This change is not acceptable. We have to make change in application.

PR #98, opened 2019-05-15
Update master with Tizen_4.0 branch

lslachciak:


lslachciak:
Travis check failed because not all commits were checked with eslint before.
We want to keep master the same with tizen 4.0 so merging now

PR #105, opened 2019-05-15
Virtual Listview: fix for missing items on list

TomaszLukawskiSam:
[Issue] N/A
[Problem] Virtual list scrolls outside of entries
Reproduction Steps:
Tizen Web UI --> Lis --> Vitual List --> Virtual List Normal Style 1 --> scroll up

[Solution] Basically issue was canvas element included to list.
Several methods in widget did not take account this element
and methods to calculate different sizes ran wrong.
The patch fix above methods.


src/js/core/widget/core/VirtualListview.js
```diff
@@ -509,6 +512,11 @@
 				self._currentIndex = index;
 			}
 
+			function getFirstElementChild(element) {
+				return (element.firstElementChild.tagName !== "CANVAS") ?
```
pkaczmarczy: Please consider changing this condition to opposite logic and return first <li> element.

src/js/core/widget/core/VirtualListview.js
```diff
@@ -652,6 +661,10 @@
 				return result;
 			}
 
+			function fnRemoveCanvas(element) {
```
pkaczmarczy: I think that different name for instance: "isNotCanvas" or "isListItem" would suit better,
Also please consider changing to opposite logic and return true for elements such as <li> or <ol>

src/js/core/widget/core/VirtualListview.js
TomaszLukawskiSam: Done

src/js/core/widget/core/VirtualListview.js
TomaszLukawskiSam: Thanks. Done.

src/js/core/widget/core/VirtualListview.js
```diff
@@ -444,7 +446,8 @@
 					i,
 					offset,
 					index,
-					isLastBuffer = false;
+					isLastBuffer = false,
+					children = filter.call(element.children, fnFilterOnlyLiElements);
```
TomaszLukawskiSam: I'll change also this function name to "isLiElement"

TomaszLukawskiSam: @pkaczmarczy 
Hi, I made some change, could you see again.

src/js/core/widget/core/VirtualListview.js
```diff
@@ -509,6 +512,11 @@
 				self._currentIndex = index;
 			}
 
+			function getFirstElementChild(element) {
+				return (element.firstElementChild.tagName === "LI") ?
+					element.firstElementChild : element.firstElementChild.nextElementSibling;
```
pkaczmarczy: Please consider returning the first element which is actually "LI" element. Second element also can be something different than "li".
This may cause problems in the future in case if listview widget structure will change.

TomaszLukawskiSam: Done.

PR #114, opened 2019-05-17
(mobile) Listview: Sometime colored listview doesn't refresh

TomaszLukawskiSam:
[Issue] N/A
[Problem] File Manager UI is differet from original app.
Background below listview is not filled.
[Solution] Before refreshing the list, the previous visible item
was not reset. This sometimes cauased issue with redraw the list.


TomaszLukawskiSam:
> Typo in commit message: cauased -->caused

Fixed.

lslachciak:
After applyin patch to modify app it works OK

Issue #141, opened 2019-06-07
"pageId is not defined" error when accessing 


lslachciak:
Reproduction steps:
Visit: https://samsung.github.io/TAU/examples/mobile/UIComponents/components/list/virtuallist.html
Click any of the listed items
Observe:
Page is changed but content is not shown.
On the error console there is an error:
```
tau][6/7/2019, 9:00:23 AM] ReferenceError: pageId is not defined
    at eval (eval at <anonymous> (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:943:14), <anonymous>:51:3)
    at eval (<anonymous>)
    at Array.<anonymous> (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:943:14)
    at Object.batchCall (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:969:22)
    at Object.util.importEvaluateAndAppendElement (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:1037:10)
    at PageContainer.prototype._include (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:12859:18)
    at PageContainer.prototype.change (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:12704:28)
    at Object.routePage.open (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:46231:25)
    at Object.resolve (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:13570:13)
    at Router._loadSuccess (https://samsung.github.io/TAU/examples/mobile/UIComponents/lib/tau/mobile/js/tau.js:14239:15)
```

TomaszLukawskiSam:
Fix has been merged.

PR #146, opened 2019-06-10
Util: fix for execute inline scripts read from page content

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/141
[Problem] "pageId is not defined" error when accessing.
    - inline scripts aren't executed
[Solution] Method to include html content was changed
    - inline scripts are removed from loaded content
      and appended to document separately


lslachciak:
I'll be able to do final verification after merging to master

Issue #158, opened 2019-06-24
Cannot change slider value using keyboard


pkaczmarczy:
Slider can be selected (blue border appears), but there is no way to change it's value by keyboard

Issue #157, opened 2019-06-24
[TV] Dimmer value does not change after moving slider


pkaczmarczy:
In TV UI Components app dimmer value does not change.

PR #159, opened 2019-06-24
Add keyboard support for slider widget

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/158, https://github.com/Samsung/TAU/issues/157
[Problem] Slider should be able to implement it's own keyboard
          handling and switch to this control after beeing
          selected by KeyboardManager. "Change" event was not
          triggered after value change.
[Solution] Add possibility to delegate keyboard
           event handling to widget instance if needed.

### DIFF
--- a/src/js/core/widget/core/Slider.js
+++ b/src/js/core/widget/core/Slider.js
@@ -892,6 +892,7 @@
 				var self = this;
 
 				self._ui.barElement.classList.add(classes.SLIDER_FOCUS);
+				self.element.focus();
 			}
 
 			/**
@@ -905,6 +906,7 @@
 
 				self._ui.barElement.classList.remove(classes.SLIDER_FOCUS);
 				self._releaseWidgetFocus();
+				self.element.blur();
 			}
 
 			/**
--- a/src/js/core/widget/core/Slider.js
+++ b/src/js/core/widget/core/Slider.js
@@ -128,8 +128,9 @@
 					SLIDER_DISABLED: "ui-disabled",
 					SLIDER_HANDLER_VALUE: "ui-slider-handler-value",
 					SLIDER_HANDLER_SMALL: "ui-slider-handler-small",
-					SLIDER_FOCUS: "ui-slider-focus"
+					SLIDER_FOCUS: "ui-focus"
 				},
+				KEYBOARD_SUPPORT_EVENTS = BaseKeyboardSupport.EVENTS,
 				prototype = new BaseWidget();
 
 			Slider.prototype = prototype;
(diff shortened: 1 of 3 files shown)


src/js/core/widget/core/Slider.js
```diff
@@ -715,88 +723,138 @@
...
+			 * @protected
+			 */
 			prototype._increaseValue = function () {
 				var self = this;
 
 				self._setValue(self._value + (parseFloat(self.element.step) || 1));
+				events.trigger(self.element, "input");
 			};
 
-			// prototype._lockKeyboard = function () {
```
TomaszLukawskiSam: I cannot verify this change during this review.

TomaszLukawskiSam: I wonder if we should remove this code

pkaczmarczy: This code was used to lock keyboard support inside selected slider.
For now this is mechanism is handled in different way. I have added to base keyboard support key event delegation if any of the widgets after "actionEnter" will capture events.
Also disabling event delegation is now handled by BaseKeyboardSupport not by widget itself.

I think we can safely remove this code.

src/js/core/widget/core/Slider.js
```diff
@@ -611,14 +623,10 @@
...
-						// 	break;
-						// case "blur":
-						// 	self._onBlur(event);
-						// 	break;
-						case "keyup":
-							self._onKeyUp(event);
+						case KEYBOARD_SUPPORT_EVENTS.taushortkeypress:
+							self._onTauShortKeyPress(event);
+							break;
+						case KEYBOARD_SUPPORT_EVENTS.taulongkeypress:
```
TomaszLukawskiSam: This code nothing does.

src/js/core/widget/core/Slider.js
```diff
@@ -715,88 +723,138 @@
...
+			 * Decrease slider current value by one step (triggers only input event)
+			 * @method _decreaseValue
+			 * @member ns.widget.core.Slider
+			 * @protected
+			 */
 			prototype._decreaseValue = function () {
 				var self = this;
 
 				self._setValue(self._value - (parseFloat(self.element.step) || 1));
+				events.trigger(self.element, "input");
```
TomaszLukawskiSam: I'm afraid that this event may cause looping in app eg. smart-things. This should be checked.

TomaszLukawskiSam (changes requested):

src/js/core/widget/core/Slider.js
pkaczmarczy: I have removed this code.

pkaczmarczy: This lines have been removed.

pkaczmarczy:
Thank you for your review :)
I have updated the PR, could please do a review once again?

lslachciak:
Because currently we are not considering this feature, could you close this PR. We will reopen it when needed.

lslachciak:
After discussing with @pkaczmarczy I'm closing this issue.
We will reopen and improve when this feature required

Issue #167, opened 2019-06-28
Bundle is not called for all profiles


pkaczmarczy:
Bundle task during build is run only for the first profile.

TomaszLukawskiSam:
PR has already merged.

PR #168, opened 2019-06-28
Fix running bundle for each profile

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/167
[Problem] Bundle was called only for the first given profile.
          Task was interrupted because result of correct run
          was passed like an error.
[Solution] Fix incorrect call and change series to parallel call
           to speed up the process.


TomaszLukawskiSam:
Looks good to me

PR #169, opened 2019-06-28
Unify examples html formatting

pkaczmarczy:


pkaczmarczy:
This change should be done to tau_1.0 branch first

Issue #171, opened 2019-07-01
Usage of deprecated customElements registration APIU


pkaczmarczy:
TAU use deprecated customElements registration API (customElemensts.js:125)

TomaszLukawskiSam:
PR has already merged.

PR #172, opened 2019-07-01
Remove unused custom elements module

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/171
[Problem] Custom elements module used deprecated Custom Elements API.
[Solution] Because the module was only included to build and unused
           it was disabled to prevent from errors after API removal.


pkaczmarczy:
> How about removing
> /core/widget/customElements
> we can always restore it from git history

Done

PR #175, opened 2019-07-02
Unify html files in UIComponents app for mobile and wearable profile

pkaczmarczy:


lslachciak:
please make sure that removal of data-build-remove attribute is safe

pkaczmarczy:
> please make sure that removal of data-build-remove attribute is safe

Thank you for comment.

I have restored the data-build-remove attribute only in mobile samples (it was not used in wearable).
I did not found any usage of this attribute, but it is present in almost every code sample in almost every boilerplate. 

I think we can leave this attribute for now and check if any other tool is not using it and then remove it.

lslachciak:
OK, so please create TAU issue for 1.1 to remove data-build-remove attribute if it is safe to do it

PR #204, opened 2019-07-11
Edit radio-button template to enable it in DE

korneliakobiela:
Issue: https://github.com/Samsung/TAU-Design-Editor/issues/126
Problem:
	- In DE we need to append #main id to ui-page element because of
back-button working
	- radio.js use this element with its own id to make internal
script logic

Solution: Edit example to use other features than id of main element

### DIFF
--- a/examples/mobile/UIComponents/components/controls/radio.js
+++ b/examples/mobile/UIComponents/components/controls/radio.js
@@ -4,7 +4,7 @@
 	 * radios - NodeList object for radios
 	 * radioresult - Indicator for active radio
 	 */
-	var page = document.getElementById("radio-demo"),
+	var page = document.querySelector(".ui-page"),
 		radios = document.querySelectorAll("input[name='radio-choice']"),
 		radioresult = document.querySelector((".radio-result")),
 		idx;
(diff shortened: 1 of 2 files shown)


examples/mobile/UIComponents/components/controls/radio.js
```diff
@@ -4,7 +4,7 @@
 	 * radios - NodeList object for radios
 	 * radioresult - Indicator for active radio
 	 */
-	var page = document.getElementById("radio-demo"),
+	var page = document.querySelector(".ui-page"),
```
pkaczmarczy: In case if app has started from index.html page (main UI Components app page) there are a few containers with .ui-page class. We cannot use this selector here.

korneliakobiela: Ok. But in this situation (radio.js file) isn't it launched only for radio.html file?

pkaczmarczy: Yes, radio.js file used only with this page, but TAU only appends active page to document. Page is removed only in case if is loaded from external document and it is removed from history stack.

image

korneliakobiela:
I close this PR because of different problem solution.

Issue #207, opened 2019-07-16
TAU UI Components: Group index doesn't work after select each dependent options


TomaszLukawskiSam:
[PRECONDITION]
Install TAU UI Components app version 5.0

[STEPS]
1. Open app

2. List > Selectable List

3. Click on checkbox of Group index 2 > uncheck Group index 2

4. Click all List Item > Obs

[OBSERVE]
Group Index doesn't check after select all dependent checkboxes.

[EXPECTED RESULT]
Group Index should be selected when all dependent checkboxes are selected, and GroupIndex should be deselect when not all of dependent checkboxes is unchecked.

lslachciak:
works well

PR #208, opened 2019-07-16
UIComponents: GroupIndex has changed if all dependent checkbox are selected

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/207
[Problem] Group index doesn't work after select each dependent options
[Solution] The fix is part of UI Components application code.
In previous version was implemented use case when a user select group index checkbox.
This patch make connection also between each checkbox and group index.
If any checkbox will be unchecked then group index also will be unchecked.

### DIFF
--- a/examples/mobile/UIComponents/components/list/list-select-check.js
+++ b/examples/mobile/UIComponents/components/list/list-select-check.js
@@ -7,13 +7,16 @@
 	function handleChange(e) {
 		var checkbox = e.target,
 			checked = e.target.checked,
-			parent = tau.util.selectors.getClosestByTag(checkbox, "li"), // get parent li element
+			currentLi = tau.util.selectors.getClosestByTag(checkbox, "li"), // get parent li element
 			next = null,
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam:
@pkaczmarczy 
Could you double check my PR? Thanks.

examples/mobile/UIComponents/components/list/list-select-check.js
```diff
@@ -27,6 +30,41 @@
...
+				if (allSelected) {
+					// find the input element in li element
+					childCheckbox = previous.querySelector("input[type=checkbox]");
+					allSelected = childCheckbox.checked;
+				}
+				// go to previous element
+				previous = previous.previousElementSibling;
+			}
+			if (previous.classList.contains("ui-group-index")) {
+				// store group index chekcbox to variable
```
pkaczmarczy: typo: chekcbox -> checkbox

Issue #216, opened 2019-07-17
UIComponents: Popup not display text after select all


TomaszLukawskiSam:
[STEP]

1. Open app
2. Tap on List
3. Tap on Select List
4. Tap on select all
5. Tap on select again
6. Observe

[OBSERVE]
Popup not display text after select all

[EXPECTED RESULT]
Should display text on popup

Wrong:
![image](https://user-images.githubusercontent.com/29534410/61370489-cd9fbc00-a892-11e9-888b-a4312df4e6f2.png)

for all items selected
![image](https://user-images.githubusercontent.com/29534410/61370519-dd1f0500-a892-11e9-8c86-92dc0f1c9ace.png)

TomaszLukawskiSam:
After patch:
![image](https://user-images.githubusercontent.com/29534410/61429490-cf1ac400-a926-11e9-918d-947404afdbd0.png)

![image](https://user-images.githubusercontent.com/29534410/61429533-012c2600-a927-11e9-93c6-b1858ff2b01d.png)
![image](https://user-images.githubusercontent.com/29534410/61429562-1e60f480-a927-11e9-9af9-7cbac2d01d1f.png)

lslachciak:
works fine. Thank you

PR #217, opened 2019-07-17
(UIComponents) Fix for popup's look on the list with select mode

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/216
[Problem] UIComponents: Popup not display text after select all
[Solution] Css selectors for listview were stronger then app styles.
Selectors in the app have been made stronger.


wszczepanski97:
works good :+1:

PR #224, opened 2019-07-22
(wearable) ToggleSlider has changed look

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU-Design-Editor/issues/192
[Problem] Toggle Slider in wearable profile looks similar to toggle switch
[Solution] According to latest guidlines, ToggleSwitch should looks
 like ToggleSlider. Instead of adding different look the old look of
 ToggleSwitch has been changed.

### DIFF
--- a/examples/wearable/UIComponents/contents/controls/toggle.html
+++ b/examples/wearable/UIComponents/contents/controls/toggle.html
@@ -29,7 +29,7 @@
 					<label>
 						Bluetooth
 						<div class="ui-toggleswitch">
-							<input class="ui-switch-input" type="checkbox" />
+							<input class="ui-switch-input" type="checkbox" checked />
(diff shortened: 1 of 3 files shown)


pkaczmarczy:
1. The border color of switch is darker than background in guide
2. Disabled checked toggle looks different
3. Pressed toggle looks different
image

TomaszLukawskiSam:
> I have found two issues:
> 
>     1. Checked and disabled property are lost (by default first two toggles should be checked and second one should be disabled in IComponents/contents/controls/toggle.html)
> 
>     2. Disabled toggle (after forcing state) should look differently

Done.

TomaszLukawskiSam:
@pkaczmarczy 
Could I ask you to review my PR again? Thank you.

PR #244, opened 2019-07-25
Don't allow to go back outside tau history

lslachciak:
[Issue] https://github.com/Samsung/TAU-Design-Editor/issues/175
[Problem] Pressing Set button causes opening previous page in web browser
[Solution] Before going back check if this is starting url for tau.

### DIFF
--- a/src/js/core/history.js
+++ b/src/js/core/history.js
@@ -46,6 +46,14 @@
 					activeState: null,
 
 					/**
+					 * Property contains starting url for tau instance
+					 * @property {Object} startURL
+					 * @static
+					 * @member ns.history
+					 */
+					startURL: null,
+
+					/**
 					 * This method replaces or pushes state to history.
 					 * @method replace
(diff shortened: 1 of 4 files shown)


src/js/core/history.js
```diff
@@ -45,6 +45,14 @@
 					 */
 					activeState: null,
 
+					/**
+					 * Property contains starting url for tau instance
+					 * @property {Object} baseUrl
+					 * @static
+					 * @member ns.history
+					 */
+					baseUrl: null,
```
TomaszLukawskiSam: I verified patch. It works good but only one thing referring to property name.
The "baseUrl" may be little confusing because of TAU adds `<base>` tag to DOM. Developer may thinks that this property refers to base tag. Could you change name to something like "initialUrl", "startingUrl"?

lslachciak: changed baseUrl --> startURL

Issue #239, opened 2019-07-25
Marquee gradient effect remains even if not necessary


TomaszLukawskiSam:
Text in list item fits to width if item is selected

![image](https://user-images.githubusercontent.com/29534410/61848631-06511e00-aeaf-11e9-8d1e-2ccf06ec87fd.png)

but when list item is moving down the gradient effect is visible and remains.
![image](https://user-images.githubusercontent.com/29534410/61848652-21239280-aeaf-11e9-8ec7-12f5d2e3f384.png)

PR #246, opened 2019-07-26
Fix for remaining gradient effect in Marquee widget

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/239
[Problem] Marquee gradient effect remains even if not necessary
[Solution] There was occured several issues in Animation module.
- marquee animation was continued despite the animation stop
- in transform method was dividing by zero this causes NaN value

### DIFF
--- a/src/js/core/util/animation/animation.js
+++ b/src/js/core/util/animation/animation.js
@@ -214,7 +214,7 @@
 				copiedArgs = [].slice.call(args);
 
 				if (config) {
-					if (config.loop) {
+					if (config.loop && config.duration > 0) {
 					// when animation is in loop then we create callback on animation and to restart animation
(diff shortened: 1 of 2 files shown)


src/css/profile/wearable/changeable/theme-circle/arclistview.less
```diff
@@ -103,6 +103,10 @@
 					}
 				}
 			}
+			.ui-arc-listview-text-content {
+				overflow: hidden;
+				text-overflow: ellipsis;
```
pkaczmarczy: This causes problems when marquee is defined by user (As in UIComponents/list/MarqueeList oe SnapMarqueeList)
image

Issue #250, opened 2019-07-28
Contextual Popup: Wrong popup widht


TomaszLukawskiSam:
![image](https://user-images.githubusercontent.com/29534410/62003892-32bea180-b11e-11e9-95ca-8f1cb0a4754d.png)

Contextual popup has to fit to screen.

pkaczmarczy:
PR has been merged

PR #251, opened 2019-07-28
Contextual Popup: fix for popup width

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/250
[Problem] Contextual Popup has wrong widht
[Solution] The contextual popup width was inherit from standard popup.
 This was issue. Width "auto" has added to the contextual popup.


TomaszLukawskiSam:
After patch:
image

image

Issue #260, opened 2019-07-30
ArcListView: Highlight incorrect height


pkaczmarczy:
Highlight has incorrect height if item contains more than 2 lines of subtext.

UIComponents app screenshot:
![image](https://user-images.githubusercontent.com/38784474/62138064-4f5a0580-b2e7-11e9-95c5-0573f89544e5.png)
list -> list with sub text -> 3 line subtext

pkaczmarczy:
PR has been merged.

PR #261, opened 2019-07-30
Fix Arclistview highlight height

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/260
[Problem] Height was set only once - before listitem height
          was set to its final value.
[Solution] Add transitionend listener for each modified
           style property


TomaszLukawskiSam:
In my opinion commit title is not adequate to the issue.

Issue #266, opened 2019-08-01
UIComponents: Button with icon in header is not visible


TomaszLukawskiSam:
UIComponents/components/headerfooter/header/lefticon.html

![image](https://user-images.githubusercontent.com/29534410/62281949-fc02c700-b44e-11e9-97fe-0042099088da.png)

lslachciak:
@TomaszLukawskiSam could you share repro path?

TomaszLukawskiSam:
The issue is not occurring in TAU 1.0

lslachciak:
1.0 https://samsung.github.io/TAU/examples/1.0/examples/mobile/UIComponents/components/headerfooter/header/lefticon.html
1.1 https://samsung.github.io/TAU/examples/1.1/examples/mobile/UIComponents/components/headerfooter/header/lefticon.html

TomaszLukawskiSam:
The issue is caused by patches dedicated for DE. Purpose of the change was add feature to buttons to change icons and other properties in during life of widget. In short, this change was required for change button look without DE page reload.

PR #288, opened 2019-08-04
Button: Fix for css classes indicating on using of icon in button

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/266
[Problem] UIComponents: Button with icon in header is not visible
[Solution]
Several issue with setting of widget option were fixed.
- fix for wrong condition for textContent,
- fix for wrong setting for size option
- fix for wrong inline option recognition
Added new option "enabledIcon".
Added method for recognition common attributes eg. "ui-inline"
Added method for widget specific attributes like "ui-btn-icon"

image

### DIFF
--- a/src/js/core/widget/BaseWidget.js
+++ b/src/js/core/widget/BaseWidget.js
@@ -214,6 +214,9 @@
 				TYPE_FUNCTION = "function",
 				disableClass = "ui-state-disabled",
 				ariaDisabled = "aria-disabled",
+				commonClasses = {
+					INLINE: "ui-inline"
+				},
 				__callbacks;
 
 			BaseWidget.classes = {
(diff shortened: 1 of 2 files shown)


src/js/core/widget/BaseWidget.js
```diff
@@ -810,27 +820,44 @@
...
+						this.options[name] = element.classList.contains(className);
+						// property exists in classname
+						return true;
+					}
 				}
 
 				return false;
 			};
 
+			prototype._readCommonOptionFromElementClassname = function (element, name) {
```
pkaczmarczy: Please add method description.

TomaszLukawskiSam: The description has added.

Issue #275, opened 2019-08-01
DropDownMenu: Press effect remains after select


TomaszLukawskiSam:
UIComponents/components/controls/dropdownmenu.html

![image](https://user-images.githubusercontent.com/29534410/62283851-c233bf80-b452-11e9-8338-d326eb262756.png)

lslachciak:
Not reproducible on 1.0
https://samsung.github.io/TAU/examples/1.0/examples/mobile/UIComponents/components/controls/dropdownmenu.html

TomaszLukawskiSam:
The patch has already merged.

lslachciak:
fixed

PR #294, opened 2019-08-06
DropDownMenu: disable focus when keboardSupport is disabled

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/275
[Problem] DropDownMenu: Press effect remains after select
[Solution] Check keyboardSupport before focus set

### DIFF
--- a/src/js/profile/mobile/widget/mobile/DropdownMenu.js
+++ b/src/js/profile/mobile/widget/mobile/DropdownMenu.js
@@ -263,7 +263,7 @@
 					native: "ui-dropdownmenu-native",
 					top: "ui-dropdownmenu-options-top",
 					bottom: "ui-dropdownmenu-options-bottom",
-					focus: "ui-focus"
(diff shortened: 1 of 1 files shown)


src/js/profile/mobile/config.js
```diff
@@ -34,6 +34,7 @@
 			ns.setConfig("popupTransition", "slideup", true);
 			ns.setConfig("pageTransition", "slide", true);
 			ns.setConfig("enablePageScroll", false, true);
+			ns.setConfig("keyboardSupport", false);
```
pkaczmarczy: This line is not needed. The common config exists in: src/js/core/config.js. KeyboradSupport is disabled by default.

TomaszLukawskiSam: Done. Thanks.

pkaczmarczy:
Please check my previous comment. After last push the change disappeared.

PR #264, opened 2019-07-31
Fix navigation bar - back button not working

lslachciak:
[Issue] Samsung/TAU-Design-Editor#166
[Problem] Back button not working for panels
[Solution] Generate new tauback event in history.back() which is handled by panel changer

PR #310, opened 2019-08-09
Fix navigation bar - back button not working #264

lslachciak:
[Issue] https://github.com/Samsung/TAU-Design-Editor/issues/166
[Problem] Back button not working for panels
[Solution] Generate new tauback event in history.back() which is handled by panel changer


TomaszLukawskiSam:
This patch has already merged.

Issue #286, opened 2019-08-01
MoreOptions: JS exception on more option show.


TomaszLukawskiSam:
JS exception during more option show.

![image](https://user-images.githubusercontent.com/29534410/62293573-2feae600-b469-11e9-901a-de4c029ec520.png)

lslachciak:
@TomaszLukawskiSam  do you mean this item
https://samsung.github.io/TAU/examples/1.1/examples/mobile/UIComponents/components/popup/moremenufooter.html#?popup=true
could you recheck?

TomaszLukawskiSam:
Issue still exists.

![image](https://user-images.githubusercontent.com/29534410/62439080-c4e02e80-b74a-11e9-8531-bda99fd44e48.png)

TomaszLukawskiSam:
> @TomaszLukawskiSam do you mean this item
> https://samsung.github.io/TAU/examples/1.1/examples/mobile/UIComponents/components/popup/moremenufooter.html#?popup=true
> could you recheck?

You attached example to mobile profile. Did you verified mobile profile or wearable profile?

TomaszLukawskiSam:
@lslachciak 
The issue is occurring only on wearable profile.

lslachciak:
@TomaszLukawskiSam this issue is still not reproducible.
After visiting url
https://samsung.github.io/TAU/examples/1.1/examples/wearable/UIComponents/contents/moreoptions/index.html
no issue with more options
![image](https://user-images.githubusercontent.com/2246331/62704269-3c33ed80-b9eb-11e9-9931-5de9bd609581.png)

TomaszLukawskiSam:
The issue is complex. Example of "More menu" for wearable has two different html for both circular and square profile on the same page. This cause several issues.

TomaszLukawskiSam:
@lslachciak 
Issue exists and is too complex to resolve now

PR #323, opened 2019-08-11
ArcListview: fix for JS issues in empty listview

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/286
[Problem] MoreOptions: JS exception on more option show
[Solution] The issue occured because list was empty.
 Widget was not ready for case with empty list.
 Added new conditions to prevent js issues.


TomaszLukawskiSam:
The PR doesn't solve the cause of issue. The patch has to be more complexity.

Issue #267, opened 2019-08-01
Tabs: Tab underline seems too low


TomaszLukawskiSam:
UIComponents/components/navigationelements/tabs/tabs_with_title_2.html

![image](https://user-images.githubusercontent.com/29534410/62282257-85b29480-b44f-11e9-8848-dc94d4e52f5c.png)

lslachciak:
reproduced on https://samsung.github.io/TAU/examples/1.0/examples/mobile/UIComponents/components/navigationelements/tabs/tabs_with_title_2.html

TomaszLukawskiSam:
The patch has already merged

PR #324, opened 2019-08-11
Tabbar: fix for wrong position of Tab underline

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/267
[Problem] Tabs: Tab underline seems too low
[Solution] Css for text in span element inside tab with title
 has been changed.


TomaszLukawskiSam:
> The commit fixes the problem and looks good but it's addressed to incorrect branch.
> Please move this commit to 1.0 branch.

Done.

Issue #222, opened 2019-07-18
Button style -> circle in header is rectangle. 


korneliakobiela:
## Description
- When the button with predefined styles (circle) is added to the application header, then a button shows as the rectangle.
## Steps to reproduce:
- Start a new project in WATT
- Mobile - Base Application
- Add Button to the application header
- Set attributes:
    - Style: circle
![image](https://user-images.githubusercontent.com/18300603/61470690-cc00f180-a981-11e9-9a83-9309ef0bf926.png)

lslachciak:
@korneliakobiela is this TAU issue?

korneliakobiela:
I think yes, because TAU is responsible for buttons style and look. But if you would like, I can transfer this issue to DE repository.

TomaszLukawskiSam:
The PR with patch is ready to review.

Sample view:
![image](https://user-images.githubusercontent.com/29534410/62832445-6c1f0300-bc2e-11e9-8196-4e4a32268df1.png)

PR #325, opened 2019-08-11
Header: add possibility to make button with circle style

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/222
[Problem] Button style -> circle in header is rectangle
[Solution] The header didn't allow that button has circle style,
 the patch makes this possible.

### DIFF
--- a/src/css/profile/mobile/changeable/common/header.less
+++ b/src/css/profile/mobile/changeable/common/header.less
@@ -87,6 +87,15 @@
 		font-stretch: condensed;
 		border-radius: 0;
 
+		&.ui-btn-circle {
+			border-radius: 50%;
+			&.ui-btn-active {
+				&::before {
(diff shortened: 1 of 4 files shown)


src/css/profile/mobile/changeable/common/header.less
```diff
@@ -87,6 +87,15 @@
 		font-stretch: condensed;
 		border-radius: 0;
 
+		&.ui-btn-circle {
+			border-radius: 52 * @unit_base;
```
pkaczmarczy: Please consider setting it to 50%. It could be the source of the problems in case of dimensions change or TV profile.

TomaszLukawskiSam:
@pkaczmarczy 
Hi, I would ask you to check my PR again. Thank you.

PR #327, opened 2019-08-13
(wearable) ToggleSlider has changed look

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU-Design-Editor/issues/192
[Problem] Toggle Slider in wearable profile looks similar to toggle switch
[Solution] ToggleSwitch should looks like ToggleSlider.
 Instead of adding different look the old look of
 ToggleSwitch has been changed.


TomaszLukawskiSam:
ok

PR #351, opened 2019-09-02
engine: add getBinding instance search on widget wrapper

TomaszLukawskiSam:
[Issue] Refer to https://github.com/Samsung/TAU-Design-Editor/issues/272
[Problem] When wiget has wrapper developer cannot find widget instance by getBinding method.
[Solution] Added possibility to find widget instance by widget wrapper.

### DIFF
--- a/src/js/core/engine.js
+++ b/src/js/core/engine.js
@@ -125,6 +125,14 @@
 				 */
 				DATA_BOUND = "data-tau-bound",
 				/**
+				 * @property {string} [DATA_WIDGET_WRAPPER="data-tau-wrapper"] attribute informs that widget has wrapper
+				 * @private
+				 * @static
+				 * @readonly
+				 * @member ns.engine
+				 */
+				DATA_WIDGET_WRAPPER = "data-tau-wrapper",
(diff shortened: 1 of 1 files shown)


src/js/core/engine.js
```diff
@@ -317,6 +335,15 @@
 
 					if (binding && typeof binding === "object") {
 						return getInstanceByElement(binding, element, type);
+					} else {
+						// Check if widget has wrapper
+						if (element && typeof element.hasAttribute === TYPE_FUNCTION &&
+								element.hasAttribute(DATA_WIDGET_WRAPPER)) {
+							wrapper = slice.call(element.children).filter(filterBuiltWidget)[0];
```
pkaczmarczy: I think it's not appropriate name. You are searching for the actual widget base element not a wrapper.
Shouldn't we care about elements that may be located deeper?

TomaszLukawskiSam: Thank you for review. I purposely assumed that the base element is no deeper than on the first level. I don't think so base element can be located deeper. We should avoid this.

PR #360, opened 2019-09-04
(test) fix for old test based old API - not existent "ej" instance

TomaszLukawskiSam:
[Issue] N/A
[Problem] Not existent alias of TAU namespace cause test fail
[Solution] Added "ej" alias for TAU framework


TomaszLukawskiSam:
This PR is not needed. Can be close.

PR #361, opened 2019-09-04
(tct) Change version of TCT to 5.5

TomaszLukawskiSam:
[Issue] N/A
[Problem] The TCT are building for previous 5.0 version
[Solution] Modification of Tizen version in tasks


TomaszLukawskiSam:
> commit title is wrong (should be 5.5 instead of 5.0

Done. Thank You.

Issue #366, opened 2019-09-06
SnapList: Newly added items cannot be selected


pkaczmarczy:
New items after refresh take correct syling, but they cannot be selected

pkaczmarczy:
This issue is caused by incorrect scrollable height calculation during refresh.
This should be recalculated during refresh function. This is a quite complicated task when listview is already built because some elemens are changing its dimensions (scaling to show hide effect)

lslachciak:
I'll close this issue after getting confirmation that it is working well

PR #371, opened 2019-09-09
SnapListview: scroll issue

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/342
[Problem] ui-listview-divider breaks listview
[Solution] Adding snap points definition to module of scrolling

Signed-off-by: Tomasz Lukawski <t.lukawski@samsung.com>

PR #372, opened 2019-09-09
Set SnapListview scrolling height during refresh

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/366
[Problem] During list refresh height was never changed.
          Newly added elements which did not fit into
          scrollable area could not be selected.
[Solution] Set scrollable area height during widget refresh


pkaczmarczy:
This PR can be closed. The same problem is resolved in https://github.com/Samsung/TAU/pull/371

Issue #368, opened 2019-09-06
Unknown 'speechSynthesis' error in listview-only page


martijnvankekem:
After opening the app, for a very specific group of people (I haven't found out which people do have this problem and which people don't. A friend of mine does have this problem while I have the same watch and don't have this problem), the app reports an error immediately on the following:

Uncaught TypeError: Cannot assign to read only property 'speechSynthesis' of object '#'
filename: file:///lib/tau/wearable/js/tau.js
line: 2195
column: 23

The page only contains a listview and there is nothing in my code calling anything close to a speechSynthesis property.

lslachciak:
We have this on to do list. We need to consult web engine team for this issue too.
Maybe this issue is related somehow to accessibility support (speech synthesis) in web engine

lslachciak:
@martijnvankekem could you attach sample app so we can try reproduce this issue?

pkaczmarczy:
I assume that this problem is caused by fastMerge function. For tau 1.1.3 dist the line which throws an exception corresponds with this piece of code:
```
/**
 * Attach fields from second object to first object.
 * @method fastMerge
 * @param {Object} newObject
 * @param {Object} orgObject
 * @return {Object}
 * @static
 * @member ns.util.object
 */
fastMerge: function (newObject, orgObject) {
	var key;

	for (key in orgObject) {
		if (orgObject.hasOwnProperty(key)) {
			newObject[key] = orgObject[key];
		}
	}
	return newObject;
},
```

I have forced problem reproduction by calling ```tau.util.object.fastMerge(window, {speechSynthesis: {}});``` (this only reproduces error code, not the actual application behavior) and I get:
```
tau.js:2199 Uncaught TypeError: Cannot assign to read only property 'speechSynthesis' of object '#<Window>'
    at Object.fastMerge (tau.js:2199)
    at <anonymous>:1:17
```
We could add additional checks to merge functions (it would help, and may prevent further problems), but I still don't know why this method has been called.

martijnvankekem:
I also don't know why this function would be called, and unfortunately, I can't reproduce where it came from, because I received this error message from users that are using my app. 

But, because this function returns the object, isn't this just a matter of returning a modified copy of 'newObject'?

So something like:

`modObject = {};
for (var attrname in newObject) { modObject[attrname] = newObject[attrname]; }
return modObject`

pkaczmarczy:
[...]

PR #379, opened 2019-09-12
Add writable property check for object merge method

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/368
[Problem] In case of attempt to override read only property
          of original object exception would be thrown.
[Solution] Add check that would protect against such situation.

### DIFF
--- a/src/js/core/util/object.js
+++ b/src/js/core/util/object.js
@@ -57,11 +57,17 @@
 				 * @member ns.util.object
 				 */
 				fastMerge: function (newObject, orgObject) {
-					var key;
+					var key,
+						propertyDescriptor;
 
 					for (key in orgObject) {
 						if (orgObject.hasOwnProperty(key)) {
-							newObject[key] = orgObject[key];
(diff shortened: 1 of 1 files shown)


martijnvankekem (approved):

lslachciak (commented):
I think that we should output to console message that changing of read only property has occurred

pkaczmarczy:
I also did some performance testing and for 100.000 calls I get about 14% (from 720 ms to 820 total) slower execution time - object with 7 properties each, 4 of them should be overridden in merge.

During loading UIComponents merge object methods are called about 20-25 times.

Merge functions are also used in gesture module for creating event data. But there is no noticeable performance drop.

I attach the histograms of measured time for gesture event usage (yaxis range is different due to number of samples used in experiment).

Figure_1

The histograms look similar, there is significant shift between them across xaxis.

TomaszLukawskiSam (approved):
Looks good to me.

Issue #369, opened 2019-09-06
Option menu icons disappearing on Gear S3.


martijnvankekem:
When using the 'More options circle' widget on a wearable device, the icons disappear upon selecting, only on the Gear S3 (as far as I found out).

The reason behind this is that the transform: scale(1) property when selected is replaced by transform: scale(0, 8236), while it should be replaced with scale(0.8236) in order to apply the correct scale.

I was able to 'hack' a fix by adding a .replace("scale(0, 8236)", "scale(0.8236)") in the code, but wasn't able to find out the reason of this happening on the Gear S3.

It's not in all versions, but at least in 1.0.3 and I think it was also in the latest 1.1 release.

lslachciak:
We have this on to do list.

lslachciak:
@martijnvankekem could you attach sample app so we can try reproduce this issue?

martijnvankekem:
I was able to fix this issue with the following 'hack', which of course isn't something you'd want in a production version:

1. I changed the ITEM_NORMAL_SCALE to 0.8 for simplicity
2. I added these lines below the `[obj].style.transform.replace(DEFAULT.ITEM_ACTIVE_SCALE,
							DEFAULT.ITEM_NORMAL_SCALE);` in the ns.widget.wearable.Selector._setActiveItem function and also below the `newTransformStyle = transform.replace(DEFAULT.ITEM_NORMAL_SCALE,
						DEFAULT.ITEM_ACTIVE_SCALE);` line.

`active.style.transform = active.style.transform.replace("scale(0, 8)", "scale(0.8)");`
`active.style.transform = active.style.transform.replace("scale(0,8)", "scale(0.8)");`

TomaszLukawskiSam:
I cannot reproduce the issue. I try to understand what happen. In my opinion possible are two options:
1.The issue is caused by device localization and  a changing number format from "0.8" to "0,8" in web engine.
2. CSS "scale()" property allows one or two arguments. Maybe web engine converted this style argument from one argument to two arguments eg. "scale(0.8)" to "scale(0, 8)"

TomaszLukawskiSam:
@martijnvankekem 
I prepared patch (https://github.com/Samsung/TAU/pull/389) for a test.

ampanasiuk:
Reproducible on Galaxy Watch. Thanks to @martijnvankekem for debugging the root cause. I've tweaked the solution a bit in #408.

TomaszLukawskiSam:
[...]

PR #389, opened 2019-09-15
(wearable) Selector: fix for scale active item

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/369
[Problem] Option menu icons disappearing on Gear S3
[Solution] Is possible that some devices convert number values
 into string using number format specific for localization region.
 The fix prevents to use comma for decimal places.

### DIFF
--- a/src/js/profile/wearable/widget/wearable/Selector.js
+++ b/src/js/profile/wearable/widget/wearable/Selector.js
@@ -334,7 +334,7 @@
 					ITEM_RADIUS: -1,
 					ITEM_START_DEGREE: 30,
 					ITEM_END_DEGREE: 330,
-					ITEM_NORMAL_SCALE: "scale(" + STATIC.SCALE_FACTOR + ")",
+					ITEM_NORMAL_SCALE: "scale(" + (STATIC.SCALE_FACTOR).toString().replace(",", ".") + ")",
 					ITEM_ACTIVE_SCALE: "scale(1)",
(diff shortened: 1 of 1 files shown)


martijnvankekem:
As the OP, I'll check tonight if this fix worked, but because It's basically the same as my 'hack', just executed prettier, I think it will work. You'll hear from me!

martijnvankekem (changes requested):
Unfortunately, this doesn't work. Hang tight, because I did figure out what the problem is and how it could/should be solved

The following things were discovered by me during testing.
1. The string itself, that is concatenated, DOES contain the right decimal place.
2. The problem is not in the concatenation of the strings, it goes wrong upon requesting the current transform tag. When you set the element's scale manually to `0.8235` in the console, and after that request the transform tag by entering `[element].style.transform`, the decimal place IS replaced.
3. This problem ONLY occurs when your watch is using a display language that uses a comma for its decimal places (for example, when my watch is set to English, this problem doesn't occur, but is does when it's set to Dutch).

With other words, the `.replace` tag isn't doing anything because `scale(0.8235)` isn't found in the `[element].style.transform`, because that string contains the `scale(0,8235)`.

martijnvankekem:
See #391

TomaszLukawskiSam:
> See #391

I'm during verification of your solution for this issue.

TomaszLukawskiSam:
The patch is ineffective.

Issue #387, opened 2019-09-13
NumberPicker: auto-created 'Set' button not destroyed when calling destroy()


martijnvankekem:
When calling the `destroy()` method of a NumberPicker, the auto-created 'Set' button does get its class `ui-number-picker-set` removed, but the element itself stays in place.

This goes wrong when executing these steps:

1. A page is called by `tau.changePage("#page");`
2. The NumberPicker is created
3. An other page is called by `tau.changePage("#otherPage");` and thus the `NumberPicker.destroy()` gets called.
4. The page is called again by `tau.changePage("#page");`
5. The NumberPicker is created again, with another new button, so now there are 2 buttons visible.

TomaszLukawskiSam:
The patch has already been merged.

PR #390, opened 2019-09-15
(wearable) Number, Time and Date pickers: remove "set" button on widget destroy

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/387
[Problem] NumberPicker: auto-created 'Set' button not destroyed
 when calling destroy()
[Solution] Method destroy has been changed for all pickers.
 After this change, if the "set" button was created by widget
 then the button will be removed in destroy method.

### DIFF
--- a/src/js/profile/wearable/widget/wearable/DatePicker.js
+++ b/src/js/profile/wearable/widget/wearable/DatePicker.js
@@ -536,8 +536,10 @@
 			 * @protected
 			 */
 			prototype._destroy = function () {
-				this._unbindEvents();
-				this.element.innerHTML = "";
+				var self = this;
+
+				self._unbindEvents();
+				self.element.innerHTML = "";
 			};
 
 			/**
--- a/src/js/profile/wearable/widget/wearable/NumberPicker.js
+++ b/src/js/profile/wearable/widget/wearable/NumberPicker.js
@@ -126,6 +126,15 @@
 					label: null,
 					footer: null
 				};
+
+				// property contains information which DOM elements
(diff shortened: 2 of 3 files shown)


TomaszLukawskiSam:
@pkaczmarczy 
Hi, Pawel could you review my PR? Thanks.

src/js/profile/wearable/widget/wearable/DatePicker.js
```diff
@@ -130,6 +130,14 @@
 				self._circleIndicatorSupporter = null;
 				self._activeSelector = null;
 				self._spins = {};
+
+				// property contains information which DOM elements
+				// was built during widget build process.
+				// These data are needed in destroy method
+				self._wasBuilt = {
```
pkaczmarczy: As I see all properties of this object would be always set to false.
However widget "set" button is placed inside <footer> which is a child of `div.ui-date-picker' (`this.element`), and its inner HTML is cleared in destroy method.

Currently this code has no effect (but I think it would be safer to put footer inside page instead of `div.ui-date-picker`)

Please consider removing this code, or moving footer and its button into page.

pkaczmarczy (changes requested):

src/js/profile/wearable/widget/wearable/DatePicker.js
TomaszLukawskiSam: Done.

pkaczmarczy (approved):
Looks good to me

PR #391, opened 2019-09-15
(wearable) Selector: fix for scale active item

martijnvankekem:
[Issue] #369
[Problem] Upon retrieval of the string after calling .style.transform, the decimal place is replaced by a comma if the watch is set to a display language that uses commas for decimal places.
[Solution] Replace the decimal place comma by a dot.

Improvement (or actually working solution) for #389

### DIFF
--- a/src/js/profile/wearable/widget/wearable/Selector.js
+++ b/src/js/profile/wearable/widget/wearable/Selector.js
@@ -946,13 +946,13 @@
 				if (active) {
 					active.style.transform =
 						active.style.transform.replace(DEFAULT.ITEM_ACTIVE_SCALE,
-							DEFAULT.ITEM_NORMAL_SCALE);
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam (approved):
The patch doesn't looks beautiful but works good. I can accept this changes as hot fix but the issue has to be fix more agile.

TomaszLukawskiSam (changes requested):
Can you organize your PR and complete commit message?

martijnvankekem:
@TomaszLukawskiSam what do you want me to organize exactly?

TomaszLukawskiSam:
> @TomaszLukawskiSam what do you want me to organize exactly?

Squash the commits to one commit with nice commit message like other commits.

martijnvankekem:
See #408, a much better solution, didn't think of that.

PR #394, opened 2019-09-17
[!fixup] SnapListview: scroll issue

HunseopJeong:
[Problem] Scrolling sometimes freezes in SnapListview
[Solution] Initialize the current index when initializing
           the SnapListview

### DIFF
--- a/src/js/core/util/scrolling.js
+++ b/src/js/core/util/scrolling.js
@@ -885,6 +885,16 @@
 				return scrollingElement === element;
 			}
 
+			/**
+			 * Update current index position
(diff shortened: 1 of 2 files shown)


HunseopJeong:
@TomaszLukawskiSam,
When repeatedly entering the SnapListview page in the UIComponent Sample app, 
I checked that the problem is caused by the current index is not being initialized.
I solved it by adding an API that sets the current index value, but if there is a better way, plz recommend it.

HunseopJeong:
This PR is duplicated with #395.
So I closed PR.

PR #395, opened 2019-09-17
Scrolling: Reset current index for snap point list

TomaszLukawskiSam:
[Issue] N/A
[Problem] Bezel scrolling not working on next list
[Solution] If next list has a less items than previous
 there was issue with snap point list. The current index is reseting
 after this patch.


singa2000:
After more checking, we'll upload it again.

singa2000 (approved):
Good

lslachciak (approved):

PR #396, opened 2019-09-19
(wearable) SnapListview: reverting change of column layout in LI item

TomaszLukawskiSam:
[Issue] N/A
[Problem] Snaplitview layout is broken
[Solution] The change of layout direction for LI item in SnapListview
 has been reverted. The column layout has changed only for ArcListview.


TomaszLukawskiSam:
@singa2000 
The patch is ready for review but we are try to verify it on device.

lslachciak (approved):

Issue #397, opened 2019-09-19
Progress element cannot be selected in Design Area


korneliakobiela:
## Technical
Profile: Mobile
## Description
User cannot select added progress element in the Design Area. Only the structure view is available then.
## Steps to reproduce
1. Create a new project in Design Editor
2. Add progress element
3. Save the file
4. Close and Reopen DE (until Samsung/TAU-Design-Editor#368 is done).
5. Try to select progress element clicking on it

## Expected behaviour
The progress element is selected.
## Actual behaviour
Content element is selected.

pkaczmarczy:
@korneliakobiela Could you please check this issue, I think it has been already fixed.

lslachciak:
@korneliakobiela please reopen issue if problem occurs
Approriate PR was merged

korneliakobiela:
It's ok, the problem doesn't exist.

PR #398, opened 2019-09-19
Enabling pointer events for progress circle

korneliakobiela:
Issue: #397
Problem: In Design Editor Progress element is not draggable item because
of none pointer-events
Solution: Remove `pointer-events: none` from ui-progress-circle and add
it to its children

### DIFF
--- a/src/css/profile/mobile/changeable/common/progress.less
+++ b/src/css/profile/mobile/changeable/common/progress.less
@@ -97,9 +97,10 @@ tau-progress {
 .ui-progress-circle {
 	position: relative;
 	display: inline-block;
-	pointer-events: none;
 	box-sizing: border-box;
-
+	* {
+		pointer-events: none;
+	}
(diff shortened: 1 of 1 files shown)


lslachciak:
@TomaszLukawskiSam could you review?

src/css/profile/mobile/changeable/common/progress.less
```diff
@@ -97,9 +97,10 @@ tau-progress {
 .ui-progress-circle {
 	position: relative;
 	display: inline-block;
-	pointer-events: none;
 	box-sizing: border-box;
-
+	* {
```
TomaszLukawskiSam, line 101: This change make sense.
`Elements with pointer-events set to none will be ignored, and the element below it will be returned.`
But I have to check possible negative effects.

TomaszLukawskiSam (approved):
Verified. Looks good to me.

PR #399, opened 2019-09-20
Prevent to touch screen on rotary event

HunseopJeong:
[Issue] N/A
[Problem] Scrolling with the wheel causes a problem because
the touch is pressed but not terminated
[Solution] Forces touch termination logic when scrolling using wheel

### DIFF
--- a/src/js/core/util/scrolling.js
+++ b/src/js/core/util/scrolling.js
@@ -393,6 +393,11 @@
 
 				previousIndex = currentIndex;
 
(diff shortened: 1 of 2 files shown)


TomaszLukawskiSam (approved):
Looks good to me but I cannot verified this patch on device.

lslachciak (approved):

singa2000 (approved):
LGTM

PR #402, opened 2019-09-22
Skip rotary event when the element is not visible

HunseopJeong:
[Issue] N/A
[Problem] Even when the screen is not visible, rotary event is applied
[Solution] Skip the rotary event when the element is not visible

### DIFF
--- a/src/js/core/util/scrolling.js
+++ b/src/js/core/util/scrolling.js
@@ -385,11 +385,25 @@
 			}
 
 			/**
+			 * Check visible state of the element
+			 * @param {Element} elm
+			 */
+			function _isVisible(elm) {
+				var rect = elm.getBoundingClientRect();
+
+				return direction ? rect.width : rect.height;
+			}
+
+			/**
 			 * Handler for rotary event
 			 * @param {Event} event
 			 */
 			function rotary(event) {
 				var eventDirection = event.detail && event.detail.direction;
+
+				if (scrollingElement && !_isVisible(scrollingElement)) {
+					return;
+				}
 
 				previousIndex = currentIndex;
 


src/js/core/util/scrolling.js
```diff
@@ -384,11 +384,32 @@
 				}
 			}
 
+			/**
+			 * Check visible state of the element
+			 * @param {Element} elm
+			 */
+			function _isVisible(elm) {
+				const style = getComputedStyle(elm);
```
TomaszLukawskiSam: In this case for checking the visibility of element enough elm.getBoundingClientRect() and check height property.

src/js/core/util/scrolling.js
```diff
@@ -384,11 +384,32 @@
...
 			/**
 			 * Handler for rotary event
 			 * @param {Event} event
 			 */
 			function rotary(event) {
+				if (scrollingElement && !_isVisible(scrollingElement)) {
+					return;
+				}
+
 				var eventDirection = event.detail && event.detail.direction;
```
TomaszLukawskiSam: move `var eventDirection;` above condition.

TomaszLukawskiSam (changes requested):

TomaszLukawskiSam (approved):
Verified. Works good.

src/js/core/util/scrolling.js
```diff
@@ -384,13 +384,26 @@
 				}
 			}
 
+			/**
+			 * Check visible state of the element
+			 * @param {Element} elm
+			 */
+			function _isVisible(elm) {
+				var rect = elm.getBoundingClientRect();
+				return direction ? rect.width : rect.height;
```
TomaszLukawskiSam, line 394: Add blank line after var (eslint)

TomaszLukawskiSam (changes requested):

TomaszLukawskiSam (approved):

lslachciak (approved):

PR #408, opened 2019-09-26
[Wearable] Fix selector icons disappearing

ampanasiuk:
[Issue] https://github.com/Samsung/TAU/issues/369
[Problem] updating style.transform with single-arg scale() value in js
          makes the decimal point replaced with comma, at least on
          some locales
[Solution] use two-arg style.transform scale() value

### DIFF
--- a/src/js/profile/wearable/widget/wearable/Selector.js
+++ b/src/js/profile/wearable/widget/wearable/Selector.js
@@ -336,7 +336,6 @@
 					ITEM_END_DEGREE: 330,
 					ITEM_NORMAL_SCALE: "scale(" + STATIC.SCALE_FACTOR + ")",
 					ITEM_ACTIVE_SCALE: "scale(1)",
-					ITEM_MOVED_SCALE: "scale(0.92)",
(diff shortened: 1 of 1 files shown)


ampanasiuk:
This works in that icons don't disappear, but sadly retrieving style.transform has "scale(0,8235, 0,8235)" and this doesn't work with "newTransformStyle = transform.replace(DEFAULT.ITEM_NORMAL_SCALE, DEFAULT.ITEM_ACTIVE_SCALE);", so the active item's size is not increased.

ampanasiuk:
New approach

TomaszLukawskiSam (changes requested):
Hi, These changes you should do in `src/js/profile/wearable/widget/wearable/Selector.js`
The `dist` folder is replaced during build process.

TomaszLukawskiSam (approved):

TomaszLukawskiSam (changes requested):
Hi, would you change the PR to branch `tau_1.0`?

TomaszLukawskiSam (approved):

Issue #409, opened 2019-09-26
Wearable UIComponents: "Open timepicker" blue background


pkaczmarczy:
![W-5 0-circle-x86-2019-09-26-123725](https://user-images.githubusercontent.com/38784474/65682096-8a17ba00-e05a-11e9-9d9c-32349280dba5.png)

The background should not be visible

pkaczmarczy:
PR has been already merged

PR #410, opened 2019-09-26
Add missing btn-text class for button

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/409
[Problem] Blue highlight was visible if button was active
[Solution] Add missing ui-btn-text only class for button


TomaszLukawskiSam (approved):
Looks good to me.

Issue #411, opened 2019-09-27
ExpandableList active effect is applied to whole sublist


pkaczmarczy:
As in description.
Example screen:
![Screenshot from 2019-09-27 10-34-51](https://user-images.githubusercontent.com/38784474/65754992-7f206080-e112-11e9-8860-be2910ec1f05.png)

pkaczmarczy:
PR has been already merged.

PR #412, opened 2019-09-27
Expandalbe list: apply active state only for it's header

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/411
[Problem] The whole expandable section was
          highlighted when clicked
[Solution] Change styling to highlight only header element


TomaszLukawskiSam (approved):
Verified. Looks good to me.

Issue #413, opened 2019-09-27
SwipeList: Text and icon are overlapped


pkaczmarczy:
​​Icon overlaps text after activating swipe list element
![wc-0927-1-2019-09-27-123924](https://user-images.githubusercontent.com/38784474/65765987-f2819c80-e129-11e9-8107-5ff9e9f0291c.png)

pkaczmarczy:
PR has been already merged.

PR #414, opened 2019-09-27
Change positioning of swipelist item text and icon

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/413
[Problem] Text and icon were overlapping each other
          (the problem existed only in wearable emulator)
[Solution] Because both elements were absolute positioned,
           so use `left` property instead `margin-left` to set position.


TomaszLukawskiSam (approved):
Verified. Looks good to me.

Issue #415, opened 2019-09-29
UIComponents: Can't scroll down to the bottom of screen after back


TomaszLukawskiSam:
[PRECONDITION]
- Install app TAU UI Components

[STEP]
1. Open TAU UI Components
2. Header and Footer
3. Tap on Bottom Button
4. Back key

[OBSERVE]
Can't scroll down to the bottom of screen after back Bottom Button

[EXPECTED RESULT]
Should scroll nornal

![image](https://user-images.githubusercontent.com/29534410/65826661-7aca8380-e289-11e9-825a-c396d313498c.png)

TomaszLukawskiSam:
The PR is waiting for review.

TomaszLukawskiSam:
The patch has already been merged.

PR #416, opened 2019-09-29
UIComponents: fix for duplicate page ID in bottom button example

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/415
[Problem] UIComponents: Can't scroll down to the bottom of screen
 after back
[Solution]
All pages in app have to be uniquely ID. Conflict of page ids make
 issues with routing and page destroy method.


pkaczmarczy (approved):
Verified. It fixes the problem.

Issue #417, opened 2019-09-29
Page: position of scroll is resetting after back from another page


TomaszLukawskiSam:
[PRECONDITION]
- Install app TAU UI Components

[STEP]
1. Open TAU UI Components
2. Navigation Elements
3. Tabs
4. Blue example
5. Back key
6. Obs

[OBSERVE]
Page position is wrong after back from previous page

[EXPECTED RESULT]
Page should be positioned correct after back from previous page

TomaszLukawskiSam:
The patch has already been merged.

PR #418, opened 2019-09-29
Page: storing and restoring page scroll position on page change

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/417
[Problem] Position of scroll is resetting after back from previous page
[Solution] Position of page scroll is stored before page hide and restored
 before page show.


pkaczmarczy (approved):
Verified. Looks good to me

Issue #419, opened 2019-09-29
Highlight: There are 2 animations when tapping "x" icon


TomaszLukawskiSam:
There are 2 animations ( blue rectangle & blue circle) when tapping "x" icon

[PRECONDITION]
- Install app TAU UI Components

[STEP]
1. Open app > recommendation> text input with button>  input text >  tap on X icon> obs

2. Open app > control> tewcontrol/Text Enveloper >  input text >  tap on X icon> obs

3. any option that can be input text and tap on X icon

[OBSERVE]
 There are 2 animations ( blue rectangle & blue circle) when tapping "x" icon

[EXPECTED RESULT]

should display only blue circle

![image](https://user-images.githubusercontent.com/29534410/65831068-a1061880-e2b5-11e9-8394-6b95ce1243dc.png)

TomaszLukawskiSam:
The patch has already been merged.

PR #420, opened 2019-09-29
Highlight: fix for two items highlights in the same time

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/419
[Problem] There are 2 animations ( blue rectangle & blue circle)
 when tapping "x" icon
[Solution]
 When the button is already highlighted then parent LI element
 is not searched and highlighted.


TomaszLukawskiSam:
Result:
image

pkaczmarczy (approved):
Verified. Looks good to me

PR #438, opened 2019-10-04
(examples) UIComponents: Duplicated ID of elements causes change page issue

TomaszLukawskiSam:
[Issue] N/A
[Problem] Back button doesn't work on several pages
[Solution] Duplicated ID of page elements cause issue in router


lslachciak:
This is not rfixing the change page issue but PR is OK to avoid unexpected issues

lslachciak (approved):

Issue #405, opened 2019-09-25
Date Picker - day indicator is overlapped


pkaczmarczy:
![Screenshot from 2019-09-25 14-37-15](https://user-images.githubusercontent.com/38784474/65601236-fd112a00-dfa1-11e9-9ee1-4a7e3e58543a.png)

Day indicator is hidden under "Set" button

TomaszLukawskiSam:
The patch has already been merged.

pkaczmarczy:
The issue has been fixed.

PR #443, opened 2019-10-06
(examples) UIComponents: fix for header height in DatePicker sample

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/405
[Problem] Date Picker - day indicator is overlapped
[Solution] Developer can set diffrent header size, in this case
 good solution is set class "ui-header-small".


pkaczmarczy (approved):
Looks good to me

Issue #406, opened 2019-09-25
Time Picker - Only Hours can be changed with rotarydetent event


pkaczmarczy:
Only hours picker can be turned using bezel.
If Hours picker is not currently set as active it is always set to "01" after rotarydetent event

TomaszLukawskiSam:
The patch has already been merged.

pkaczmarczy:
The issue has been fixed.

PR #444, opened 2019-10-06
(core) BaseWidget: fix for converting number options during widget refresh

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/406
[Problem] Time Picker - Only Hours can be changed with rotarydetent event
[Solution] Bug in refresh method of base widget caused wrong running of
 time picker because of number value of "to" was converted to string.
 Additionally was fixed header size of TimePicker example and disabled
 `active` style of Hours on widget start.

### DIFF
--- a/examples/wearable/UIComponents/contents/controls/time-picker.html
+++ b/examples/wearable/UIComponents/contents/controls/time-picker.html
@@ -18,7 +18,7 @@
 
 <body>
 	<div class="ui-page" data-enable-page-scroll="false" id="number-picker-page">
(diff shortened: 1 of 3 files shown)


pkaczmarczy (approved):
Looks good to me, but I have found another issue:

Right after building time picker nothing is selected and bezel does not affect anything (that's OK).
After selecting one of the spinners bezel can spin each one correctly, but then if no spinner would be selected rotarydetent events are still handled and inactive spinner changes its value. (I will create another issue)

PR #448, opened 2019-10-09
[Wearable] Replace "change" event with "input" for Slider

grzegorz-czajkowski:
[Issue] N/A

[Problem] Mobile profile successfully emits both events
	"input" and "change" on Slider widget whereas Wearable profile
	emits only "change".

[Solution] Emits "input" event instead of "change" since according to [1]
	input event should be fired immediately when the slider is adjusted but
	change event does not fire immediately, the value does not change until
	the slider stops moving
	This fixes SThing.js behaviour with Slider and Dimmer synchronization
	for Wearable.

[Remarks] May cause regression for tau apps already using "change" event.
	To fix this we should provide real implementation for "change" event.

[1] https://www.impressivewebs.com/onchange-vs-oninput-for-range-sliders/

### DIFF
--- a/examples/wearable/UIComponents/contents/controls/slider/extended_slider.js
+++ b/examples/wearable/UIComponents/contents/controls/slider/extended_slider.js
@@ -25,7 +25,7 @@
 	 */
 	pageBeforeShowHandler = function () {
 		slider = tau.widget.Slider(elSlider);
-		elSlider.addEventListener("change", changeOpacity, false);
+		elSlider.addEventListener("input", changeOpacity, false);
 	};
 
 	/**
@@ -34,7 +34,7 @@
 	 */
 	pageHideHandler = function () {
 		slider.destroy();
-		elSlider.removeEventListener("change", changeOpacity, false);
+		elSlider.removeEventListener("input", changeOpacity, false);
 	};
 
 	page.addEventListener("pagebeforeshow", pageBeforeShowHandler, false);
--- a/src/js/profile/wearable/widget/wearable/Slider.js
+++ b/src/js/profile/wearable/widget/wearable/Slider.js
@@ -65,11 +65,11 @@
 				prototype = new CoreSlider(),
 				eventType = {
 					/**
-					 * Triggered when the section is changed.
-					 * @event change
+					 * Triggered when the slider value is changed.
+					 * @event input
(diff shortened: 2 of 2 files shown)


lslachciak (changes requested):
@grzegorz-czajkowski please make sure that CHANGE is not used by UIComponents.
After check, I'll approve

pkaczmarczy:
The change event is used in tau 1.1 wearable dimmer sample.
I think we should trigger both events.

IMO change event should be triggered after user release the slider handle and value changes, and input event should be triggered with each handle movement.

grzegorz-czajkowski:
@lslachciak 
"change" event is being used by
examples/wearable/UIComponents/contents/controls/slider/extended_slider.js
in both tau 1.0 and 1.1.
I will fix this.

@pkaczmarczy 
Where I can find source of  tau 1.1 wearable dimmer sample?
>>  I think we should trigger both events.
They behave slightly different, if someone needs "change" event we should implemented it according to spec.

lslachciak (approved):

lslachciak:
> 
> 
> The change event is used in tau 1.1 wearable dimmer sample.
> I think we should trigger both events.
> 
> IMO change event should be triggered after user release the slider handle and value changes, and input event should be triggered with each handle movement.


@pkaczmarczy 
I think we can create new issue for change event implementation in future version of TAU not blocking this one. OK?

pkaczmarczy:
Yes, I think we should fix event emitting for slider.
@grzegorz-czajkowski The dimmer sample is available only on tau_1.1 branch and file path is: examples/wearable/UIComponents/contents/controls/dimmer.js

grzegorz-czajkowski:
@pkaczmarczy Ok thanks. I will prepare separate PR to fix this.

grzegorz-czajkowski:
@lslachciak  and @pkaczmarczy 
Created issue to add "change" event - https://github.com/Samsung/TAU/issues/453
Patch to tau 1.1 is here https://github.com/Samsung/TAU/pull/449

TomaszLukawskiSam (changes requested):
In my opinion issue #453 is blocking for this PR

grzegorz-czajkowski:
This needs improvements and is out of scope currently.

Issue #450, opened 2019-10-10
UIComponents: Back the top of page when click back


TomaszLukawskiSam:
TAU version 1.1.6

[PRECONDITION]
- Install App TAU UI Components

[STEP]
1. Open app TAU UI Components
2.  Select any option =>> back => Obs

[OBSERVE]

Back the top page when click back

[EXPECTED RESULT]
Should be back to the screen before

TomaszLukawskiSam:
The patch has already been merged

PR #451, opened 2019-10-10
PageContainer: fix for options object delivered to the change method

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/450
[Problem] UIComponents: Back the top of page when click back
[Solution] The change method allows to provide object of options,
 but when object contains url property this cause confusion in routing


pkaczmarczy (approved):
Verified.

PR #459, opened 2019-10-22
Invoke init method with scrolled element

lslachciak:
[Issue] N/A
[Problem]
Section changer was not working properly after restoring app.
[Reason]
When app with section changer was restored,
refresh method was called invoking init without scrolled element
[Solution]
Call init with scrolled element as argument.


lslachciak:
@pkaczmarczy could you review?

pkaczmarczy (approved):
Verified. Looks good to me

PR #461, opened 2019-10-23
(wearable) Spin: added widget example for using in public api

TomaszLukawskiSam:
[Issue] N/A
[Problem] Pickers delivered by TAU are not enough for developers.
 Developers need more customizable widget similar to pickers.
[Solution]
 Spin widget was created for internal usage as common widget for
 NumberPicker, DatePicker, TimePicker. This commit added example
 how to use Spin widget stand alone.

### DIFF
--- a/examples/wearable/UIComponents/contents/controls/index.html
+++ b/examples/wearable/UIComponents/contents/controls/index.html
@@ -91,6 +91,11 @@
 					</a>
 				</li>
 				<li>
+					<a href="spin/index.html">
+						Spin
+					</a>
+				</li>
+				<li>
 					<a href="numberpicker/index.html">
 						Number Picker
 					</a>
--- a/examples/wearable/UIComponents/contents/controls/spin/custom-spin.html
+++ b/examples/wearable/UIComponents/contents/controls/spin/custom-spin.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html>
+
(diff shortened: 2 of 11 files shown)


lslachciak:
I got sth like
image

-100 is different than 9?
Also displaying .value doesn't seem to be needed

lslachciak:
default text is not needed I think.
Also change is not doing anything
image

lslachciak:
Also I have issue - I can't rotate items in spin group
Reporduction
1. Enter sping group
2. Rotate spin
3. Click change
4. Go backl toi the previous menu
5. Enter spin group again
6. Try to rotate spin
Observe: not able to rotate

lslachciak:
two more issues (not a blocking one)
1. if you rotate bezel without clicking spin control I'm getting warning on console
file:///contents/controls/spin/spin-group.html (110) :Uncaught TypeError: Cannot read property 'option' of undefined
2. after choosing Spin->default there is warning on console:

file:///dist/wearable/js/tau.js (18126) :Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.

lslachciak:
Also small issue
After entering custom -->spin
image
and pressing upper right virtual button

image

lslachciak (approved):

PR #466, opened 2019-11-04
_restoreContentStyle - restore only non-empty values

hsiwkin:
[Problem] _restoreContentStyle was resettig Pages' height with empty
value
[Solution] add check verifying if value that is being restored is not
empty


pkaczmarczy (approved):
Looks good to me

TomaszLukawskiSam (approved):
Looks good to me.

Issue #467, opened 2019-11-12
TAU Footer OK button -  font size increased


lmslachciak:

pkaczmarczy:
PR has been merged.

PR #468, opened 2019-11-12
Remove text scale effect on press in footer button

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/467
[Problem] Text in footer button changed its size on press
[Solution] Remove styles responsible for changing dimensions
           on press for footer button


TomaszLukawskiSam (approved):
Looks good to me.

Issue #469, opened 2019-11-12
TimePicker doesn't return time


TomaszLukawskiSam:
TimePicker should returns Date object with selected time just like to DatePicker do.

The issue was reported on StackOverflow
https://stackoverflow.com/questions/58768395/how-to-retreive-time-from-tizen-timepicker-or-is-there-a-better-way-to-do-it-wi

lslachciak:
PR merged

PR #470, opened 2019-11-12
TimePicker: fix for widget getValue method

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/469
[Problem] The TimePicker always returns NaN value
[Solution] Added previously not implemented
 protected method _getValue


pkaczmarczy (approved):
Looks good to me

PR #471, opened 2019-11-12
Util: _loop method provides loop Time for current animation frame

TomaszLukawskiSam:
[Issue] N/A
[Problem] Progress bar is not animating on value change
[Solution] Animation callback in the progress bar
 expects time as argument but not given.
 The "_loop" method in tau.util has been changed.


pkaczmarczy (approved):
Looks good to me

PR #478, opened 2019-11-15
Define the value of the Toggle elements as a number

DohyungLim:
- Toggle elements already added have an initial value of undefined.
- Set to 0 when undefined.
- Set 1 to on and 0 to off when clicking on Toggle elements.


DohyungLim:
@TomaszLukawskiSam 
Adds a patch to tau_1.0 as well as tau_1.1 .

singa2000 (approved):
LGTM

Issue #475, opened 2019-11-14
Marquee widget reset


lmslachciak:
ui-marquee widget should
- stop when another widget is selected and reset its state
- when animation stops widget should be reset

pkaczmarczy:
This issue is under investigation. The mechanics responsible for animations is quite complex, it takes time to find the root cause of the problem.

pkaczmarczy:
PR has been already merged.

PR #479, opened 2019-11-15
Fix resetting marquee element

pkaczmarczy:
[Issue] https://github.com/Samsung/TAU/issues/475
[Problem] Marquee widget was not resetting completly.
[Solution] Remove check for render method if animation is active.
           This check should not exist, because widget that has
           paused animation may want to be redrawn at any moment.

This change will not cause regression with https://github.com/Samsung/TAU/issues/239 issue
(ListView now is destroying Marquee widget after element is no longer selected)

### DIFF
--- a/src/js/core/widget/BaseWidget.js
+++ b/src/js/core/widget/BaseWidget.js
@@ -1127,14 +1127,8 @@
 				}
 			}
 
-			function render(stateObject, element, isChild, options) {
-				var recalculate = false,
-					animation = (options) ? options.animation : null;
-
-				if (animation && !animation.active) {
-					// Animation has stopped before render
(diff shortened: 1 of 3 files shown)


lslachciak (commented):
typo in commit message:

Marquee widget was **not** stopped, but not reseted.
really "not stopped"?

TomaszLukawskiSam (approved):
Looks good to me.

pkaczmarczy:
@TomaszLukawskiSam  I have added necessary properties to mocked objects in tests due to fails.
Could you please look at it again?

TomaszLukawskiSam (approved):
Ok. Works good.

PR #493, opened 2019-11-26
Remove padding-bottom for Coverflow example

lslachciak:
[Issue] https://github.com/Samsung/TAU-Design-Editor/issues/442
[Problem] Can scroll up content of coverflow example
[Solution] Set pading-bottom CSS property to 0 for ui-content.
This property is inherited from
tau.circle.css:448
.ui-page.ui-scroll-on .ui-content {
  margin-top: 7.1875rem;
  padding-bottom: 6.25rem;
  padding-top: 0;
}
It is in order to allow scrolling app content up, because some items are not visible.
In case of coverflow app this approach is not needed because widget fills in screen.


TomaszLukawskiSam (approved):
I agree with you.

PR #506, opened 2019-12-10
Bottom button highlights out of area when clicked

DohyungLim:
- Applies only the button area with the border-radius value.


TomaszLukawskiSam (approved):
Looks good to me.

DohyungLim:
thanks. Tomasz.

PR #507, opened 2019-12-10
SnapListview: fix for detect scrollend

TomaszLukawskiSam:
[Issue] N/A
[Problem] SnapListview is triggering "scrollend" and "scrollstart" on each
 "scroll" event
[Solution] Detection of scroll end is realized by set timeout on scroll.
 When scoll event is not triggering during indicated period then
 scroll end event is triggering. The timeout was set on "zero" this
 cause issue. Now, this value is changing now to 150ms.

I verified possible regression. I didn't reproduce issue.


lslachciak (approved):
No regression found

PR #509, opened 2019-12-11
Fix problem that scrollToPosition is not working correctly

HunseopJeong:
[Issue] N/A
[Problem] SnapListview.scrollToPosition is not working
[Solution] Sync the index and position with scroller
and calculate the rendered position with the scrolled
value and add the 'skipAnimation' to skip the animation

### DIFF
--- a/src/js/core/util/scrolling.js
+++ b/src/js/core/util/scrolling.js
@@ -46,6 +46,7 @@
 				// current state of scroll position
 				scrollPosition = 0,
 				lastScrollPosition = 0,
+				baseScrollPosition = 0,
 				moveToPosition = 0,
 				lastRenderedPosition = 0,
 				lastTime = Date.now(),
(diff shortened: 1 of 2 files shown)


TomaszLukawskiSam (commented):
I'm reviewing this PR at the moment.

TomaszLukawskiSam (commented):
Works better but some issues still exists.
image

These issue can be reproduce by scroll by touch and call method scrollToPosition(n)

HunseopJeong:
@TomaszLukawskiSam,
I checked issue and fixed it.
Could you re-check the issue on your reproduced steps?

TomaszLukawskiSam:
> @TomaszLukawskiSam,
> I checked issue and fixed it.
> Could you re-check the issue on your reproduced steps?

I verified again. The issue doesn't occur anymore.
Now I'll check on device in Sample app.

TomaszLukawskiSam (approved):
I verified PR on device in sample app. Works fine.

PR #510, opened 2019-12-13
(example) ArcListview: fix for issue with broken list after back to the page

TomaszLukawskiSam:
[Issue] N/A
[Problem] ArcListview widget is broken after reopen page with list of marque
[Solution] SnapListview widget was built by mistake on an already build ArcListview.
 Unnecessary JS script has removed.

image


examples/wearable/UIComponents/contents/list/arclist_marquee_style.html
```diff
@@ -187,8 +187,6 @@ <h2 class="ui-title">
 				</li>
 			</ul>
 		</div>
-		<script src="arclist_marquee_style.js">
```
lslachciak, line 190: why not remove this JS file too?

TomaszLukawskiSam: You are right. Done.

lslachciak (approved):

PR #511, opened 2019-12-16
Fix problem that scrollToPosition is not working correctly

HunseopJeong:
[Issue] N/A
[Problem] SnapListview.scrollToPosition is not working
[Solution] Sync the index and position with scroller
and calculate the rendered position with the scrolled
value and add the 'skipAnimation' to skip the animation


TomaszLukawskiSam (approved):
Verified. Looks good to me.

PR #518, opened 2019-12-20
Marquee: fix to show left gradient in Marquee List

DohyungLim:
- "Marquee List" issue in UIComponent sample.
- Left gradient does not work when Marquee List is selected.
- Removed "return null" handling to verify normal operation.


TomaszLukawskiSam (approved):
Works very good. Thanks.

PR #522, opened 2019-12-24
fixup: Fix problem that scrollToPosition is not working correctly

HunseopJeong:
[Problem] Scroll is not working correctly when using the scrollToIndex
[Solution] Restore the default position of scroller when destroying the
snaplistview

### DIFF
--- a/src/js/profile/wearable/widget/wearable/SnapListview.js
+++ b/src/js/profile/wearable/widget/wearable/SnapListview.js
@@ -163,6 +163,7 @@
 					self._currentIndex = null;
 					self._enabled = true;
 					self._isTouched = false;
+					self._isScrollToPosition = false;
 					self._scrollEventCount = 0;
 					self._marginTop = 0;
 				},
@@ -667,7 +668,7 @@
 				self._unbindEvents();
 
 				scroller = getScrollableParent(self.element);
-				if (scroller) {
+				if (scroller && !self._isScrollToPosition) {
 					scroller.scrollTop = 0;
 				}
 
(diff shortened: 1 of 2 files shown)


src/js/profile/wearable/widget/wearable/SnapListview.js
```diff
@@ -683,6 +683,7 @@
 				self._selectedIndex = null;
 				self._currentIndex = null;
 
+				scrolling.scrollToIndex(0, 0);
```
TomaszLukawskiSam: I'm not sure this change. This patch causes unwanted effects during the transition animation between pages. User can see items from top of page during page transition.

HunseopJeong: If you change to a different page Isn't it common to go to the top of the page?
Could you give me an example of a case you are worried about?

TomaszLukawskiSam: Hi,
I prepared example for check your PR. (https://github.com/Samsung/TAU/pull/524)
You can test your commit in web browser with slow animations.
image

HunseopJeong: Thank you to share the example.
I will check it!

TomaszLukawskiSam (commented):
Now, works very good. I need to check this PR in several applications.

TomaszLukawskiSam (approved):
Looks good to me.

singa2000 (approved):
LGTM
I think we need to more check in case of skip to animation of scrollToPosition().

Issue #206, opened 2019-07-15
Marquee: Add new animation


pkaczmarczy:
Currently there is no animation syle that "wraps" text. Currenlty it is possible to side text and in next cycle it appends on it's original position. 

New style should allow to slide it from right to left starting from right edge.

PR #525, opened 2019-12-30
Marquee: fix for animation type endToEnd

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/206
[Problem] Scroll style should allow to slide it from right to left
 starting from right edge.
[Solution] Added copy of content at the end of content

### DIFF
--- a/examples/wearable/UIComponents/contents/controls/marquee/index.html
+++ b/examples/wearable/UIComponents/contents/controls/marquee/index.html
@@ -14,7 +14,7 @@
 	</script>
 	<script src="../../../lib/tau/wearable/js/tau.js">
 	</script>
-</head>
+	</head>
 
 <body>
 	<div class="ui-page" data-enable-page-scroll="false" id="marqueeTest">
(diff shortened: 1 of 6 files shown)


singa2000:
@TomaszLukawskiSam 
Would you check the ci error, please?
>>     at file:///home/travis/build/Samsung/TAU/tests/js/core/widget/core/Marquee/Marquee.js:427
>>     at file:///home/travis/build/Samsung/TAU/tests/js/core/widget/core/Marquee/Marquee.js:429: 'undefined' is not a function (evaluating 'animation.reset()')
>> TypeError: 'undefined' is not a function (evaluating 'animation.reset()')

examples/wearable/UIComponents/contents/controls/marquee/index.html
```diff
@@ -24,17 +24,14 @@ <h2 class="ui-title">
 			</h2>
 		</header>
 		<div class="ui-content">
-			<div class="ui-marquee ui-marquee-gradient" data-iteration="10" data-marquee-style="slide" id="marquee">
```
grzegorz-czajkowski, line 27: Why do we remove "slide" from examples?

examples/wearable/UIComponents/contents/controls/marquee/index.html
```diff
@@ -24,17 +24,14 @@ <h2 class="ui-title">
...
-				MarqueeTEST TEST message TEST for marquee
-			</div>
-			<div class="ui-marquee ui-marquee-ellipsis" data-delay="1000" data-marquee-style="scroll" id="marquee2">
```
grzegorz-czajkowski, line 30: It would be good to have an example with **ui-marquee-ellipsis** if it's still supported.

examples/wearable/UIComponents/contents/controls/marquee/index.html
```diff
@@ -24,17 +24,14 @@ <h2 class="ui-title">
...
-				MarqueeTEST TEST message TEST for marquee MarqueeTEST TEST message TEST for marquee MarqueeTEST TEST message TEST for marquee MarqueeTEST TEST message TEST for marquee
+			<div class="ui-marquee ui-marquee-gradient" data-delay="0" data-iteration="10" data-marquee-style="scroll" id="marquee">
+				Marquee SCROLL TEST TEST message TEST for marquee MarqueeTEST TEST message TEST for marquee MarqueeTEST TEST message TEST for marquee MarqueeTEST TEST message TEST for marquee
 			</div>
 			<div class="ui-marquee ui-marquee-gradient" data-delay="2000" data-iteration="infinite" data-marquee-style="alternate" id="marquee3">
```
grzegorz-czajkowski, line 30: How about enumerating ids for marquee as marque1, marguee2, marquee3 etc.

examples/wearable/UIComponents/contents/controls/marquee/index.html
```diff
@@ -24,17 +24,14 @@ <h2 class="ui-title">
...
-				MarqueeTEST TEST message (infinite)
+				Marquee ALTERNATE TEST TEST message (infinite)
 			</div>
 			<div class="ui-marquee ui-marquee-gradient" data-delay="3000" data-iteration="1" data-marquee-style="endToEnd" id="marquee4">
```
grzegorz-czajkowski, line 33: Shouldn't we have more iterations?

grzegorz-czajkowski (commented):
Works fine on emulator. Added minor comments to examples.

examples/wearable/UIComponents/contents/controls/marquee/index.html
TomaszLukawskiSam: You are right... I shouldn't remove this example... but I have to move it to bottom of page.

examples/wearable/UIComponents/contents/controls/marquee/index.html
TomaszLukawskiSam: This is example according to guideline.

examples/wearable/UIComponents/contents/controls/marquee/index.html
TomaszLukawskiSam: It doesn't matter for me but I changed it.

DohyungLim (approved):
It looks good for execute.

grzegorz-czajkowski (approved):

TomaszLukawskiSam:
On device is occurring performance issues but, I think it is not caused by my patch.
I recorded timeline when issue with stop/pause of marquee occurred. This seems rather browser issue, but I'll check in more detail.

image

gwolny:
I have tried to reproduce the performance issue with emulator and with web simulator.
Both failed to reproduce. This supports the hypothesis that the issue is not caused by the patch.

gwolny (approved):

DohyungLim:
@gwolny @TomaszLukawskiSam 
The problem is not reproduced in the simulator and emulator. It only happens on the target device.
I think that's not a problem with the patch. Because the problem already occurs before applying the patch.
Tomasz guesses the browser's problem.
Apart from the patch, this part will need to be analyzed and corrected.

PR #526, opened 2019-12-30
[ArcListView] Add highlight to item on click

grzegorz-czajkowski:
Issue: N/A
Problem: highlight on item is not shown after back
Solution: select item on click


grzegorz-czajkowski:
I am testing it on device so please hold on reviewing it.

grzegorz-czajkowski:
I changed the code to trigger both events on click:
1. unselect (the previous item)
2. select (new item).

However, on device highlight becomes higher sometimes. This can not be reproduce on Tizen Emulator. Look like different issue.

PR #527, opened 2020-01-02
SPIN is moving automatically even though it was not up/down

HunseopJeong:
[Issue] N/A
[Problem] If you scroll quickly in hours and click quickly in minutes,
it scrolls automatically
[Solution] Check the container whether it is activated or not

### DIFF
--- a/src/js/profile/wearable/widget/wearable/TimePicker.js
+++ b/src/js/profile/wearable/widget/wearable/TimePicker.js
@@ -470,9 +470,12 @@
 			 */
 			prototype._onSpinChange = function (event) {
 				var self = this,
+					spin = null,
 					parentElement = getClosestByClass(event.target, classes.CONTAINER);
 
-				if (parentElement) {
+				spin = ns.engine.getBinding(event.target);
+
+				if (spin && spin.option("enabled") && parentElement) {
 					self.value(event.detail.value);
 				}
 			};


src/js/profile/wearable/widget/wearable/TimePicker.js
```diff
@@ -472,7 +472,7 @@
 				var self = this,
 					parentElement = getClosestByClass(event.target, classes.CONTAINER);
 
-				if (parentElement) {
+				if (event.target.classList.contains(classes.ACTIVE_LABEL) && parentElement) {
```
TomaszLukawskiSam: I cannot reproduce the issue but I understand that the issue occurred.
I have suggestion to change:
```
				var self = this,
					spin = null,
					parentElement = getClosestByClass(event.target, classes.CONTAINER);

				spin = ns.engine.getBinding(event.target);
				if (spin && spin.option("enabled") && parentElement) {
					self.value(event.detail.value);
				}
```

TomaszLukawskiSam (changes requested):

src/js/profile/wearable/widget/wearable/TimePicker.js
TomaszLukawskiSam: I verified this PR on device... but unfortunately seems like doesn't work.

HunseopJeong: @TomaszLukawskiSam ,
Yes.. It seem to only work on the TimePicker correctly.

TomaszLukawskiSam (changes requested):
I thought about this amendment and changed my mind.
Fix should be implemented to the Spin widget instead TimePicker.

The Spin widget should not generate events when it is disabled.

HunseopJeong:
I agree your opinion.
I will reinvest to find the issue :(

HunseopJeong:
@TomaszLukawskiSam 
I have confirmed that the issue does not occur if the spinchange event is not triggered when the value of the spin widget is changed. I think there is a part of the app that receives the "spinchange" event and handles it separately. What do you think? Could you check the isssue with below code on MMR?
```
        // 296 line in Spin.js
	// if (triggerChangeEvent) {
	// 	ns.event.trigger(self.element, "spinchange", {
	// 		value: parseInt(self.options.value, 10),
	// 		dValue: parseInt(self.options.value, 10) - parseInt(self._prevValue, 10)
	// 	});
	// }
```

HunseopJeong:
I closed this PR becuase the issue looks to be fixed by #531.

PR #528, opened 2020-01-02
Unbind the events on TimePicker correctly

HunseopJeong:
[Issue] N/A
[Problem] "spinchange" events are not turned off when unbinding the events
[Solution] Turn off "spinchange" events when unbinding the events


TomaszLukawskiSam (approved):
Looks good to me.

PR #529, opened 2020-01-02
[ArcListView] Add highlight to item on click

grzegorz-czajkowski:
Issue: N/A

Problem: if user clicked not active element on list, then back button is pressed,
	highlight on item is not shown *sometimes*.

Cause: click event triggers 'Change' event but it may not be handled since 'ui-page-active'
	goes to new page element (sub page) and events from old page (not active) are ignored.

Solution: select item on 'pageshow' event

Remarks: on device, highlight sometimes becomes higher for a while but this is separate issue
	which is going to be handled separately.

### DIFF
--- a/src/js/profile/wearable/widget/wearable/ArcListview.js
+++ b/src/js/profile/wearable/widget/wearable/ArcListview.js
@@ -1190,6 +1190,12 @@
 				arcListviewSelection.classList.add(classes.SELECTION_SHOW);
 			}
 
+			prototype._removeHighlight = function () {
+				var selection =	this._ui.arcListviewSelection;
+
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam (commented):
I verified this PR on device. Works good. There is issue with size of highlight but this seems be also before this PR but before was no possibility to see it.

singa2000:
I verified the PR  in the Emulator, it looks good working.

grzegorz-czajkowski:
@TomaszLukawskiSam added separate commit to this PR which fixes wrong highlight dimension on device.

TomaszLukawskiSam (approved):
This patch works very good. Thanks.

PR #531, opened 2020-01-03
Spin: Event "spinchange" has stopped when widget is in disabled state

TomaszLukawskiSam:
[Issue] N/A
[Problem] SPIN is moving automatically even though it was not up/down
[Solution] Spin event is triggering by animation, this cause issue when widget
 is disabled during animation.
 Stopping the animation on widget disabling is solution for the issue


grzegorz-czajkowski (approved):
Looks good to me.

HunseopJeong (approved):
LGTM

PR #532, opened 2020-01-05
Page, ArcListview: Changing scroll position of hidden page

TomaszLukawskiSam:
[Issue] N/A
[Problem] List text is overlap with Title
Reproduction path:
1. Select 'List'
2. Touch on 'Button List' without select before
3. Back button

[Solution] Added possibility to set the last scroll position of hidden page.
 When animation of list scroll is ending then scroll position is set for
 hidden page.

### DIFF
--- a/src/js/profile/wearable/widget/wearable/ArcListview.js
+++ b/src/js/profile/wearable/widget/wearable/ArcListview.js
@@ -487,7 +487,8 @@
 					currentTime = Date.now(),
 					startTime = state.startTime,
 					deltaTime = currentTime - startTime,
-					scroll = state.scroll;
+					scroll = state.scroll,
+					pageWidget = null;
 
 				if (deltaTime >= state.duration) {
 					self._scrollAnimationEnd = true;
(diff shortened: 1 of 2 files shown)


singa2000:
It's good working.
It looks good.

singa2000 (approved):
Good working.

src/js/profile/wearable/widget/wearable/ArcListview.js
```diff
@@ -511,6 +512,10 @@
 						eventUtils.trigger(state.items[state.currentIndex].element, "selected");
 						state.toIndex = state.currentIndex;
 
+						// set last scroll position when current page is hidden
+						pageWidget = ns.engine.getBinding(self._ui.page, "Page");
+						pageWidget.setLastScrollPosition(-1 * scroll.current || 0);
```
grzegorz-czajkowski, line 517: Who sets _lastScrollPosition to 0 when page is hide?
Shouldn't we sync page's scroller.scrollTop with ArcListView's scroll.current instead?

TomaszLukawskiSam: The _lastScrollPosition is set by Page widget on "pagebeforehide" and restored on "pagebeforeshow". The ArcListview try to change scroll position but couldn't. Now, new position of last scroll position can be changed even for hidden page.

PR #533, opened 2020-01-05
Wrong radio button selection on snaplistview

TomaszLukysky:
Fix & example for test.


TomaszLukawskiSam:
@singa2000 
I don't have the opportunity right now to check this change thoroughly but seems like the patch works.

src/js/profile/wearable/widget/wearable/SnapListview.js
```diff
@@ -744,7 +744,9 @@
 
 				targetIndex = getIndexOfSnapListItem(targetListItem, listItems);
 
-				utilEvent.preventDefault(e);
+				if (e.target.tagName === "LABEL") { // prevents duplicate events on input inside label;
```
TomaszLukawskiSam, line 747: In here we need add also check if label contains input type radio. eg. e.target.querySelector("input[type='radio']")

singa2000: I think we need to more check, which is correct behavior.

TomaszLukawskiSam: > I think we need to more check, which is correct behavior.

Yes. I agree with you.

TomaszLukawskiSam:
The PR is not needed.

PR #534, opened 2020-01-06
Sync jQuery library version

HunseopJeong:
[Issue] N/A
[Problem] jQuery version is different between uncompressed and minified
[Solution] Update uncompressed jQuery version to the 2.1.1

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 dist/**
 node_modules/**
+libs/**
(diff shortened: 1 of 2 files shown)


TomaszLukawskiSam (commented):
This change looks good to me, but it can have impact on apps which use jquery from TAU

singa2000:
It's the PR for fixing  jQuery version is different between uncompressed and minified.
I think we have to make same version uncompressed and minified.
If the PR merge,  is there any problem with TAU working?
If there is no problem, how about to merge it?

lslachciak (approved):
OK I think this should have no influence on TAU

TomaszLukawskiSam (approved):
looks good to me.

PR #535, opened 2020-01-07
Wrong radio button selection on snaplistview (different behaviour)

TomaszLukawskiSam:
This patch adding selection for button also by click on list item.


TomaszLukawskiSam:
This patch is depending to #533

TomaszLukawskiSam:
The PR is not needed.

PR #536, opened 2020-01-07
Snaplistview: Fixing for selections of listitem when none listitem selected before

TomaszLukawskiSam:
The PR is depending to PR https://github.com/Samsung/TAU/pull/535

The issue with selection is occurring after fast click on second list item.
For reproduce the issue:
1. Run SnapListview & Radio example
2. Choose 3rd item and quickly 4th item
Result: 4th items doesn't make selected


TomaszLukawskiSam:
The PR is not needed.

PR #538, opened 2020-01-10
Slide marquee take the text on the right from the gradient

DohyungLim:
- Issue created from the marquee slide effect.
- The last text on the right is obscured by the gradient.
- Get the text by raising the value of translateX.

### DIFF
--- a/src/js/core/widget/core/Marquee.js
+++ b/src/js/core/widget/core/Marquee.js
@@ -232,9 +232,12 @@
 						containerWidth = stateDOM.offsetWidth,
 						textWidth = stateDOM.children[0].offsetWidth,
 						value,
+						excludeValue,
 						returnValue;
 
-					value = state * (textWidth - containerWidth);
+					// RIGHT gradient is 85% spec.
+					excludeValue = (containerWidth * 15 / 100) / 2;
+					value = state * (textWidth - containerWidth + excludeValue);
 					returnValue = "translateX(" + (-1 * round100(value) || 0) + "px)";
 					if (current === returnValue) {
 						return null;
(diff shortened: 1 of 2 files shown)


src/js/core/widget/core/Marquee.js
```diff
@@ -234,7 +234,7 @@
 						value,
 						returnValue;
 
-					value = state * (textWidth - containerWidth);
+					value = state * (textWidth - containerWidth + 25);
```
grzegorz-czajkowski: I've applied your PR and rebuild TAU. With your change I can see last character of text on right side (no cutting) in UI Components app. 
Please have in mind that this change is made to core (src/js/core/widget/core/Marquee.js) implementation and will applied for both Mobile and Wearable profiles. Such hard coded values may bring unexpected result in some cases. Can we find alternative way to fix it? For example by applying percentage etc.

grzegorz-czajkowski (changes requested):

grzegorz-czajkowski:
Don't we have the same issue for alternate style?

singa2000:
A little issue also occurs in alternate style.

src/js/core/widget/core/Marquee.js
DohyungLim: I measured the right gradient size at 15%.
and the size was divided by two and divided evenly.
I tested it on the Marquee Sample, the Snap Marquee List, and the Marquee List, and they all confirmed normal operation.

please re-check and report me.

TomaszLukawskiSam (approved):
I verified. I don't find any issues. Looks good to me.

singa2000 (commented):
It's good working for me.

singa2000:
@grzegorz-czajkowski 
Would you review the PR,please?

grzegorz-czajkowski (approved):
Thanks for taking my comment into considerations. Looks good to me.

Issue #477, opened 2019-11-14
SnapListview: Cannot build widget with one or none items


pkaczmarczy:
Exception is thrown as below:
![Screenshot from 2019-11-14 11-59-07](https://user-images.githubusercontent.com/38784474/68851499-5bbb7000-06d6-11ea-9eb4-c15b4e8e4f55.png)

Issue probably exists on 1.0 too, but I did not checked it.

martijnvankekem:
I can confirm that this bug is also present in TAU 1.0.20.

lslachciak:
@HunseopJeong could you merge PR https://github.com/Samsung/TAU/pull/540 if ready and close the issue?

HunseopJeong:
Merged the PR(#540) for this issue.

PR #540, opened 2020-01-12
Fix the problem that cannot build widget with one or none items

HunseopJeong:
[Issue] #477
[Problem] Scroller is not assigned when the item is one or none
[Solution] Assign the body element when there is no scrollable parent

### DIFF
--- a/src/js/profile/wearable/widget/wearable/SnapListview.js
+++ b/src/js/profile/wearable/widget/wearable/SnapListview.js
@@ -448,13 +448,25 @@
 				// finding page  and scroller
 				ui.page = utilSelector.getClosestByClass(listview, "ui-page") || document.body;
 
-				scroller = getScrollableParent(listview);
+				scroller = getScrollableParent(listview) ||
+					ui.page.querySelector(".ui-scroller") ||
+					ui.page;
 				if (scroller) {
+					scroller.classList.add(classes.SNAP_CONTAINER);
+					ui.scrollableParent.element = scroller;
+
+					visibleOffset = scroller.clientHeight;
+					ui.scrollableParent.height = visibleOffset;
+
+					listItem = listview.querySelector(self.options.selector);
+					if (!listItem) {
+						return scroller;
+					}
+
 					if (!scrolling.isElement(scroller)) {
 						scrolling.enable(scroller, "y");
 					}
 
-					listItem = listview.querySelector(self.options.selector);
 					elementHeight = (listItem) ? listItem.getBoundingClientRect().height : 0;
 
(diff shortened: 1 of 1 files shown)


src/js/profile/wearable/widget/wearable/SnapListview.js
```diff
@@ -448,7 +448,7 @@
 				// finding page  and scroller
 				ui.page = utilSelector.getClosestByClass(listview, "ui-page") || document.body;
 
-				scroller = getScrollableParent(listview);
+				scroller = getScrollableParent(listview) || ui.page;
```
TomaszLukawskiSam: I have suggestion to this code. The standard scrolling layer is the element with class "ui-scroller". This element is created during build process of page.
```
scroller = getScrollableParent(listview) ||
    ui.page.querySelector(".ui-scroller") ||
    ui.page;
```

TomaszLukawskiSam (changes requested):

singa2000:
In my test,  the issue  is reproduced SnapList  has just no item.
If  it has one item, there is no problem.
But after the PR, if there is no list but it can  scroll by touch and it's not good behavior.

TomaszLukawskiSam:
> In my test, the issue is reproduced SnapList has just no item.
> If it has one item, there is no problem.
> But after the PR, if there is no list but it can scroll by touch and it's not good behavior.

I agree with you. The PR didn't resolve all issue with empty snaplistview.
like: 

> Uncaught TypeError: Cannot read property 'position' of undefined
>     at getSnapPointIndexByScrollPosition (tau.js:16124)
>     at touchEndCalculateSpeed (tau.js:16168)
>     at HTMLDocument.touchEnd (tau.js:16224)
> getSnapPointIndexByScrollPosition @ tau.js:16124
> touchEndCalculateSpeed @ tau.js:16168
> touchEnd @ tau.js:16224

HunseopJeong:
@TomaszLukawskiSam,
I changed some orders to fix the error.
Please retry to review this PR.

TomaszLukawskiSam (approved):
Verified. Works good but we need also changes for a use case that developer want to add new item to list.
```
var l = tau.engine.getBinding(snaplistElement),
     li = document.createElement('li');

li.innerHTML = "aaaaaaaaaaaa";
l.element.appendChild(li);
l.refresh();
```

This can be done by another PR.
Patch for this may looks like this https://github.com/TomaszLukawskiSam/TAU/tree/add-item-to-snaplistview

HunseopJeong:
@TomaszLukawskiSam,
Okay. After merging this PR, please make a PR for the proposed use case.
I think we would be good to add a new API for the 'addItem'.

PR #541, opened 2020-01-13
SnapListview: Fix to get a correct index from the widget

HunseopJeong:
[Issue] N/A
[Problem] getSelectedIndex get an wrong index after come back from
another page
[Solution] Return the selected index

### DIFF
--- a/src/js/profile/wearable/widget/wearable/SnapListview.js
+++ b/src/js/profile/wearable/widget/wearable/SnapListview.js
@@ -729,7 +729,7 @@
 			 * @member ns.widget.wearable.SnapListview
 			 */
 			prototype.getSelectedIndex = function () {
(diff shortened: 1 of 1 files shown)


HunseopJeong:
Issue Reproduce:
1.Open page with snaplistview
2.Create snaplist widget
3.Subscribe to “selected” event and check inside getSelectedIndex
4.Do prescroll on saved index(0 by default)
5.Scroll to another position(for example 2) & update saved index
6.Go to another page and ensure that snaplist is destroyed
7.Return to prev page

after 7 step, getSelectedIndex returned same index value if you change the selected element using the scroll.

TomaszLukawskiSam (approved):
I agree. The currentIndex represents snaplistview lindex even during transition. While selectedIndex represents selected item and will be null during rotation.

PR #539, opened 2020-01-10
Enable rotary scrolling for circle wearable devices

hsiwkin:
[Problem] Rotary scrolling didn't work in circle wearable devices.
[Solution] Enable rotary scrolling in circle-helper

Signed-off-by: Hubert Siwkin <h.siwkin@samsung.com>

PR #543, opened 2020-01-13
Revert "Enable rotary scrolling for circle wearable devices"

lslachciak:
Reverts Samsung/TAU#539

As per comment:

"This patch adding redundant support for rotation for content & lists. The lists have own support for rotary event. This PR may cause issues hard to predict."


TomaszLukawskiSam (approved):
I agree.

PR #545, opened 2020-01-14
Marquee: fix infinite animation

gwolny:
[Issue] N/A
[Problem] Infinite animation is running only for one iteration.
[Solution] Change "iteration" type to "string". When using string both
number and "infinite" are parsed correctly.

Signed-off-by: Grzegorz Wolny <g.wolny@samsung.com>

### DIFF
--- a/src/js/core/widget/core/Marquee.js
+++ b/src/js/core/widget/core/Marquee.js
@@ -195,7 +195,7 @@
 				defaults = {
 					marqueeStyle: style.SLIDE,
 					speed: 60,
-					iteration: 1,
+					iteration: "1",
 					currentIteration: 1,
 					delay: 0,
 					timingFunction: "linear",
--- a/tests/js/core/widget/core/Marquee/api/marquee.js
+++ b/tests/js/core/widget/core/Marquee/api/marquee.js
@@ -1,11 +1,15 @@
+/* global equal, test */
 (function (ns) {
-    "use strict";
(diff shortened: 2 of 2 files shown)


lslachciak (changes requested):
Besides style errors reported by Travis
I think you should check if parseInt returns NaN
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

gwolny:
Before the patch if not a number string was given, it was always converted to the NaN. With the patch string "infinite" is correctly handled. Other are converted to the NaN. So for other cases it works as before.

lslachciak (approved):

lslachciak:
please wait with merge

.eslintrc.js
```diff
@@ -336,6 +336,8 @@ module.exports = {
 		"jsdoc"
 	],
 	"globals": {
-		"tizen": true
+		"tizen": true,
+		"equal": false,
```
TomaszLukawskiSam: This is not general true. The "equal" and "test" are only in tests.

Better solution is define
`/* global equal, test */` in specific file
or better modify `/*global test */` and modify test like:
```
	test("API of core Marquee Widget", function (assert) {
             assert.equal( ... );
     ...   
```

TomaszLukawskiSam (changes requested):

TomaszLukawskiSam (approved):
Looks good to me.

Issue #546, opened 2020-01-15
[UI Component] Inconsistent behaviour of Time Picker vs. Date Picker


lmslachciak:
[ How to reproduce]

UI Component App -> Controls -> Time Picker 24H -> nothing is selected

UI Component App -> Controls -> Date Picker -> month is selected

[ Problem]

Inconsistent behaviour between Time Picker and Date Picker.

Nothing should be selected in Date Picker or "Hrs" should be selected in Time Picker.

lslachciak:
Issue fixed

PR #547, opened 2020-01-15
Don't set selected month in DatePicker

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/546
[Problem] In UI Components app DatePicker example has month selected.
Other Examples like DateTime Picker have no default value selected.
[Solution] Modify DatePicker to not set default selection.


TomaszLukawskiSam (approved):
Verified. Looks good to me.

Issue #548, opened 2020-01-15
[UIComponent] Header is cut when no much items on list


korneliakobiela:
Reproduction steps

TAU 1.0.22 + UIComponents

1. Run UIComponents

2. Go to Popup

3. Rotate clockwise

Observe

Expectation:

"Popup" header gets hidden

Observation

"Popup" header stays cut

lslachciak:
Issue fixed

PR #549, opened 2020-01-15
Fix hiding header while scrolling on short lists

korneliakobiela:
[Issue] https://github.com/Samsung/TAU/issues/548
[Problem] Header isn't totally hidden while list is scrolling and there
is too few elements on the one.
[Solution] Enforce minimal list height

### DIFF
--- a/src/css/profile/wearable/changeable/theme-circle/arclistview.less
+++ b/src/css/profile/wearable/changeable/theme-circle/arclistview.less
@@ -4,7 +4,7 @@
 	overflow: hidden;
 	top: 0;
 	left: 0;
-
+	min-height: 250 * @unit_base;
 	&-carousel {
 		position: absolute;
 		top: 0;


src/css/profile/wearable/changeable/theme-circle/arclistview.less
```diff
@@ -4,7 +4,7 @@
 	overflow: hidden;
 	top: 0;
 	left: 0;
-
+	min-height: 250px;
```
lslachciak: As I see in this file all values base on the @unit_base
e.g.
width: 360 * @unit_base;
height: 108 * @unit_base;

I think we should use it here also

lslachciak (changes requested):

src/css/profile/wearable/changeable/theme-circle/arclistview.less
korneliakobiela: Ok, I will change it. Do you know if `@unit_base` is the similar to pixel?

lslachciak: Yes, I think this was intended as pixel independent

korneliakobiela: Done!

src/css/profile/wearable/changeable/theme-circle/arclistview.less
```diff
@@ -4,7 +4,7 @@
 	overflow: hidden;
 	top: 0;
 	left: 0;
-
+	min-height: 250* @unit_base;
```
lslachciak: OK with unit base, but after some additional check I think that the best place for definition is theme.layout.less file which defines similar items 
e.g.
https://github.com/Samsung/TAU/blob/tau_1.0/src/css/profile/wearable/changeable/theme-changeable/theme.layout.less
https://github.com/Samsung/TAU/blob/tau_1.0/src/css/profile/wearable/changeable/theme-circle/theme.layout.less

korneliakobiela: I think it is consistent with code-style of the rest LESS files. 
**Example** 
https://github.com/Samsung/TAU/blob/tau_1.0/src/css/profile/wearable/changeable/theme-circle/listview.less#L37

lslachciak: OK

TomaszLukawskiSam (approved):
Verified. I didn't find regression. This PR looks good to me.

lslachciak (approved):

PR #551, opened 2020-01-15
fixup! fixup! fixup! Fix problem that scrollToPosition is not working correctly

grzegorz-czajkowski:
Issue: N/A

Problem: In SnapListView, while scrolling an item gets bigger

Cause: while scrolling, _listItemAnimate(scrollValue) function is called
	where current scoll position is passed as param. When '0' (no scroll)
	is passed, the following function is called:

		 scrolling.getScrollPosition()

	which returns outdated scroll position (before scroll) during scroll (when
	scroll finishes and 'touchend' comes, scroll position is fine).

Solution:
	Accept '0' as correct position of scroll passed to _listItemAnimate() func.
	Pass scrolling.getScrollPosition() to _listItemAnimate on refresh func to
	restore scroll position.

### DIFF
--- a/src/js/profile/wearable/widget/wearable/SnapListview.js
+++ b/src/js/profile/wearable/widget/wearable/SnapListview.js
@@ -343,14 +343,11 @@
 			prototype._listItemAnimate = function (scrollValue) {
 				var self = this,
 					anim = self.options.animate,
-					animateCallback = self._callbacks[anim],
-					scrollPosition;
+					animateCallback = self._callbacks[anim];
 
 				if (animateCallback) {
-					scrollPosition = scrollValue || scrolling.getScrollPosition();
-
 					utilArray.forEach(self._listItems, function (item) {
-						item.animate(scrollPosition, animateCallback);
+						item.animate(scrollValue, animateCallback);
 					});
 				}
 			};
@@ -379,7 +376,9 @@
 					removeSelectedClass(self);
 				}
 
-				self._listItemAnimate(event && event.detail && event.detail.scrollTop);
+				if (event && event.detail) {
+					self._listItemAnimate(event.detail.scrollTop);
+				}
 
 				// scrollend handler can be run only when all touches are released.
 				if (self._isTouched === false) {
@@ -505,7 +504,7 @@
 				}));
 
 				self._listItems = listItems;
-				self._listItemAnimate();
+				self._listItemAnimate(scrolling.getScrollPosition());
 			};
 
 			/**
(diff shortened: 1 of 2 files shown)


src/js/profile/wearable/widget/wearable/SnapListview.js
```diff
@@ -343,14 +343,11 @@
...
 					anim = self.options.animate,
-					animateCallback = self._callbacks[anim],
-					scrollPosition;
+					animateCallback = self._callbacks[anim];
 
 				if (animateCallback) {
-					scrollPosition = scrollValue || scrolling.getScrollPosition();
-
 					utilArray.forEach(self._listItems, function (item) {
-						item.animate(scrollPosition, animateCallback);
```
lslachciak, line 353: this function is also called in other place
function scrollHandler(self, event) {
self._listItemAnimate(event && event.detail && event.detail.scrollTop);
we need to make sure that event will always exist in this function otherwise we will call this function with undefined scrollValue

grzegorz-czajkowski: I didn't meet such scenario but ok we can prevent from such case.

src/js/profile/wearable/widget/wearable/SnapListview.js
```diff
@@ -379,7 +376,9 @@
 					removeSelectedClass(self);
 				}
 
-				self._listItemAnimate(event && event.detail && event.detail.scrollTop);
+				if (event && event.detail) {
```
lslachciak, line 379: how about 
if (event && event.detail && event.detail.scrollTop) {
	self._listItemAnimate(event.detail.scrollTop);
}

grzegorz-czajkowski: 0 is correct values for scroll (scrollTop). We should pass it to _listItemAnimate.

lslachciak: OK

grzegorz-czajkowski:
I pushed the same change again due to Travis issue.

grzegorz-czajkowski:
>> Message: Expected 3 assertions, but 2 were run
>> at file:///home/travis/build/Samsung/TAU/tests/libs/qunit-1.11.0.js:412
>>   at runTests (file:///home/travis/build/Samsung/TAU/tests/js/profile/wearable/widget/wearable/SnapListview/SnapListview.js:204)

This was introduced in latest change. Need to check this.

HunseopJeong (commented):
It works well.
Thanks @grzegorz-czajkowski,
If you fix the CI issue, it would be good to merge it.

lslachciak (approved):
Don't see any regression with this PR. Problem is fixed. Seems OK

PR #552, opened 2020-01-16
Fix the scroll position when moving to the last index

HunseopJeong:
[Issue] N/A
[Problem] The scroll position is misaligned when scroll to the last
index
[Solution] Adjust the position when moving the last index


HunseopJeong:
@TomaszLukawskiSam,
The scroll position is misaligned when moving to the last index using the 'scrollToPositiion'.
So I tried to find the correct solution, but I didn't find the fundamental reason.
I have uploaded a hot fix, but please review it.

lslachciak:
I'm checking this PR together with https://github.com/Samsung/TAU/pull/551
And it seems no regression for SnapListView

lslachciak (approved):
I don't see regression in this patch

Issue #566, opened 2020-01-27
ArcListView - wrong selection before entering menu item


lmslachciak:
[wrong_selection.zip](https://github.com/Samsung/TAU/files/4116116/wrong_selection.zip)
Reproduction:
UIComponents --> List ->ArcListView -->ArcListView
start scrolling  + Long Tap (don' release finger)
Observe:
Wrong selection visible (see attached video)
Expected:
No selection or correct selection

lslachciak:
This is reproducible also in other cases, but the long tap has easiest reproduction steps.

grzegorz-czajkowski:
Selection is shown even if scroll is stopped by touch start. Selection should be shown when:
 * animation finished (not stopped by any touch events) or
 * touch end (if stopped by touch start while scrolling).

grzegorz-czajkowski:
Patch merged, issue can be closed.

PR #568, opened 2020-01-28
ArcListview: do not show selection while scrolling

grzegorz-czajkowski:
Issue: https://github.com/Samsung/TAU/issues/566

Problem: stopping fast scroll by pressing finger causes
	selection at wrong position.

Cause: scrollAnimationEnd variable is also set on TouchStart event
	to indicate that current animation should end.
	But in fact we are in the middle of animation.

Solution: show selection only if scroll finishes regardless of scrollAnimationEnd.

### DIFF
--- a/src/js/profile/wearable/widget/wearable/ArcListview.js
+++ b/src/js/profile/wearable/widget/wearable/ArcListview.js
@@ -511,10 +511,14 @@
 						1
 					);
 					if (self._scrollAnimationEnd) {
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam (approved):
Works and looks good. Thanks.

lslachciak (approved):

lslachciak:
works well

PR #576, opened 2020-01-30
Move variable declarations before their usage

lslachciak:
[Issue] N/A
[Problem] Sonar reported issues:
Move the declaration of "states" before this usage.
Move the declaration of "selectors" before this usage.
Move the declaration of "progressValues" before this usage.
[Solution] Move declarations

### DIFF
--- a/src/js/core/util/deferredWhen.js
+++ b/src/js/core/util/deferredWhen.js
@@ -32,6 +32,9 @@
 				var i = 0,
 					resolveValues = [].slice.call(arguments),
 					length = resolveValues.length,
+					progressValues,
+					progressContexts,
+					resolveContexts,
 
 					/**
 					 * The count of uncompleted subordinates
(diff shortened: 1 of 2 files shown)


src/js/core/util/deferredWhen.js
```diff
@@ -61,6 +61,9 @@
 					 * @member ns.util.deferredWhen
 					 * @private
 					 */
+					progressValues,
```
TomaszLukawskiSam: These variables should be moved above description of "updateFunc" ?

TomaszLukawskiSam (changes requested):

src/js/core/util/deferredWhen.js
lslachciak: done

TomaszLukawskiSam (approved):
Looks good to me.

PR #577, opened 2020-01-30
Fix potential TypeError

lslachciak:
[Issue] N/A
[Problem]
Sonar issue:
TypeError can be thrown as "values" might be null or undefined here.
[Solution] Add additional check

### DIFF
--- a/src/js/support/mobile/widget/Datetimepicker.js
+++ b/src/js/support/mobile/widget/Datetimepicker.js
@@ -427,6 +427,8 @@
 							case 4:
 								values = getCalendar().months.names.slice();
 								break;
+							default:
+								values = [];
 						}
 						if (values.length === 13) { // @TODO Lunar calendar support
 							if (values[12] === "") { // to remove lunar calendar reserved space


src/js/support/mobile/widget/Datetimepicker.js
```diff
@@ -428,7 +428,7 @@
 								values = getCalendar().months.names.slice();
 								break;
 						}
-						if (values.length === 13) { // @TODO Lunar calendar support
+						if (values && values.length === 13) { // @TODO Lunar calendar support
```
TomaszLukawskiSam: I think it is not solution of the issue.
We need "default" case in switch statement in where `values` will be assign

TomaszLukawskiSam (changes requested):

src/js/support/mobile/widget/Datetimepicker.js
lslachciak: done

TomaszLukawskiSam (approved):
Looks good to me.

PR #579, opened 2020-01-30
Remove Math and Set overrides

lslachciak:
[Issue] N/A
[Problem] Sonar reported issues:
Remove this override of "Math"
Remove this override of "Set"
[Solution]
Rename overrides


TomaszLukawskiSam (approved):
Looks good to me.

PR #587, opened 2020-02-03
(SVACE) engine.js: fix expression for element always evaluate to true

TomaszLukawskiSam:
(SVACE) engine.js: fix expression for element always evaluate to true
    
    [Issue] SVACE 560208
    [Problem] Refactor this code so that this expression does not always
     evaluate to true.
    [Solution]
     - 560208 - element has been removed from expression
    

grzegorz-czajkowski (commented):
Right. This condition for element is double checked.

grzegorz-czajkowski (approved):

PR #598, opened 2020-02-04
(SVACE) Progress: define step function

grzegorz-czajkowski:
Issue: svace 560237
Problem: make this function anonymous by removing its name
Solution: declare this function (as step) to be used internally


TomaszLukawskiSam (approved):
Verified. Works and looks good.

PR #601, opened 2020-02-05
(SVACE) Datetimepicker: Remove unnecessary return

grzegorz-czajkowski:
Issue: svace 560246
Problem: Refactor this method to not always return the same value.
Solution: Remove return statement since it's useless

### DIFF
--- a/src/js/profile/wearable/widget/wearable/Datetimepicker.js
+++ b/src/js/profile/wearable/widget/wearable/Datetimepicker.js
@@ -118,7 +118,6 @@
 			 *
 			 * @param {ns.widget.wearable.Datetimepicker} self
 			 * @param {Event} event
-			 * @return {null}
 			 */
 			function handleWheelButtons(self, event) {
 				var button = event.target,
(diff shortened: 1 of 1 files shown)


src/js/profile/wearable/widget/wearable/Datetimepicker.js
```diff
@@ -135,7 +135,7 @@
 					direction = -1;
 					// Everything else than minus / plus button should be ignored
 				} else if (!buttonClassList.contains(classes.uiDatetimeWheelPlus)) {
-					return null;
+					return;
```
TomaszLukawskiSam, line 137: Remove @return {null} from js doc for this method

TomaszLukawskiSam (changes requested):

src/js/profile/wearable/widget/wearable/Datetimepicker.js
grzegorz-czajkowski: Right. Thanks.

TomaszLukawskiSam (approved):

PR #602, opened 2020-02-05
(SVACE) Slider: added "return" to "_setValue" method

TomaszLukawskiSam:
[Issue] svace 560249
[Problem] Refactor this function to use "return" consistently.
[Solution] Added missing "return null" at the end of method

### DIFF
--- a/src/js/profile/wearable/widget/wearable/Slider.js
+++ b/src/js/profile/wearable/widget/wearable/Slider.js
@@ -527,6 +527,7 @@
 				} else {
 					return CoreSliderPrototype._setValue.call(self, value);
 				}
+				return null;
 			};
 
 			/**


src/js/profile/wearable/widget/wearable/Slider.js
```diff
@@ -527,6 +527,7 @@
 				} else {
 					return CoreSliderPrototype._setValue.call(self, value);
```
grzegorz-czajkowski, line 528: src/js/core/widget/core/Slider.js does not return anything. I think this return is useless. Please consider removing it instead of adding return at the end of the function.

TomaszLukawskiSam: I based on js doc above method. There is {number|null}

grzegorz-czajkowski: I am not sure if this function returns number. I let you make the decision about this.

TomaszLukawskiSam: I'm afraid that method `CoreSliderPrototype._setValue` could returns number. For backward compatibility with others version I prefer left this "return" statement

grzegorz-czajkowski (approved):

PR #603, opened 2020-02-05
(SVACE) performance: add return at the end of get function

grzegorz-czajkowski:
Issue: svace 560273
Problem: Refactor this function to use "return" consistently.
Solution: add return at the end of get function


TomaszLukawskiSam (approved):
Looks good to me.

PR #605, opened 2020-02-05
(SVACE) ToggleSwitch.extra: remove unreachable code

grzegorz-czajkowski:
Issue: svace 560271, 560272
Problem: Change this condition so that it does not always evaluate to "false"
Solution: remove this code due it's never called. In js, typeof null is equal to
	object and null will be handled in above condition.

### DIFF
--- a/src/js/support/mobile/widget/ToggleSwitch.extra.js
+++ b/src/js/support/mobile/widget/ToggleSwitch.extra.js
@@ -534,9 +534,6 @@
 
 					// If changes came from input value change
 				} else {
-					if (value === null) {
-						value = getInitialValue(tagName, control);
-					}
 					if (isNaN(value)) {
 						return;
 					}


src/js/support/mobile/widget/ToggleSwitch.extra.js
```diff
@@ -534,9 +534,6 @@
 
 					// If changes came from input value change
 				} else {
-					if (value === null) {
```
TomaszLukawskiSam, line 537: I'm not sure this change. We probably don't take account "undefined" value, but before also didn't take account it.

grzegorz-czajkowski: Yes, undefined was not handled and adding code for this may cause different behavior.

TomaszLukawskiSam: Ok. Looks good to me.

TomaszLukawskiSam (approved):

PR #606, opened 2020-02-05
Remove code blocks that are never executed

hsiwkin:
[Problem] Svace issue: condition in if statement  is always true
[Solution] Reorganize code so that there are no if statements that
contains expressions that are always true

### DIFF
--- a/src/js/profile/wearable/widget/wearable/SnapListview.js
+++ b/src/js/profile/wearable/widget/wearable/SnapListview.js
@@ -458,7 +458,7 @@
 						scrolling.enable(scroller, "y");
 					}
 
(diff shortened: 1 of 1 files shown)


lslachciak (approved):

grzegorz-czajkowski (approved):

TomaszLukawskiSam (changes requested):
Could you verify this PR on device?

hsiwkin:
@TomaszLukawskiSam I've checked it on a device and everything seems to work fine.

TomaszLukawskiSam (approved):
I verified on device. Everything works also in application based on TAU.

PR #607, opened 2020-02-05
(SVACE) SnapListview: move animation function out of requestAnimationFrame

hsiwkin:
(SVACE) SnapListview: move animation function out of requestAnimationFrame
    
[Problem] function declaration (and definition) inside requestAnimationFrame
[Solution] Move animation function out of requestAnimationFrame


TomaszLukawskiSam (commented):
This PR has to be verified on device. We must be sure that nothing breaks.

hsiwkin:
I've checked it on the device and it works OK

TomaszLukawskiSam:
Double verification has been done. Everything works also in application based on TAU.

TomaszLukawskiSam (approved):

PR #609, opened 2020-02-06
(SVACE) Datetimepicker: omit number conversion to string

grzegorz-czajkowski:
Issue: svace 560253, 560254, 560255, 560256, 560257, 560258, 560259, 560260
Problem: Convert this operand into a number.
Solution: let updateField() function operate on number instead of string


TomaszLukawskiSam (approved):
Looks good to me.

PR #611, opened 2020-02-06
(SVACE) SearchBar: remove unnecessary code

grzegorz-czajkowski:
Two commits:
 * (SVACE) SearchBar: remove unnecessary condition
 *  (SVACE) SearchBar: remove label element

### DIFF
--- a/src/js/support/mobile/widget/SearchBar.js
+++ b/src/js/support/mobile/widget/SearchBar.js
@@ -447,7 +447,6 @@
 					searchBox,
 					clearButton,
 					cancelButton,
-					labelDiv,
 					frontIcon,
 					label = findLabel(element),
 					searchBoxClasses,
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam (approved):
Looks good to me.

TomaszLukawskiSam (changes requested):
I need to verify it.

src/js/support/mobile/widget/SearchBar.js
```diff
@@ -662,17 +654,9 @@
 			function inputBlur(self) {
 				var ui = self._ui,
 					inputtedText = ui.input.value,
```
TomaszLukawskiSam: 'inputtedText' is assigned a value but never used

TomaszLukawskiSam (changes requested):

src/js/support/mobile/widget/SearchBar.js
grzegorz-czajkowski: Good catch. Thanks.

TomaszLukawskiSam (approved):

PR #618, opened 2020-02-10
Add/coverage api

hsiwkin:


package.json
```diff
@@ -5,6 +5,7 @@
   "version": "1.0.25",
   "scripts": {
     "test": "grunt test",
+    "test:karma": "karma start tests/karma/single.conf.js",
```
TomaszLukawskiSam: use local karma package instead of global package
./node_modules/karma/bin/karma

package.json
```diff
@@ -44,9 +45,10 @@
     "grunt-qunit-tap": "~2.0.0",
     "grunt-string-replace": "~0.2.4",
     "grunt-sync": "^0.6.2",
-    "karma": "^1.3.0",
+    "karma": "^4.4.1",
```
TomaszLukawskiSam: are you sure we need update karma version?

tests/karma/single.conf.js
```diff
@@ -48,7 +48,11 @@ module.exports = function (config) {
 			{pattern: "tests/js/**/*.html", included: false, served: true, watch: true},
 			{pattern: "src/js/**/*.js", included: false, served: true, watch: true},
 			{pattern: "dist/**/*", included: false, served: true, watch: true},
-			"tests/karma/single.test.path.js",
+			{pattern: "libs/BezierCurve.js", included: false, served: true},
```
TomaszLukawskiSam: Instead of modify this file we can update `all.conf.js`

tests/karma/single.conf.js
```diff
@@ -85,11 +89,12 @@ module.exports = function (config) {
 			}
 		},
 
-		browsers: ["bigScreen"],
+		// browsers: ["bigScreen"],
+		browsers: ["ChromeHeadless"],
```
TomaszLukawskiSam: I think a better solution is change `bigScreen.base` to `Chrome`

TomaszLukawskiSam (changes requested):

package.json
hsiwkin: I've restored the "old" version of Karma

tests/karma/single.conf.js
hsiwkin: done

tests/karma/all.conf.js
```diff
@@ -50,6 +50,7 @@ module.exports = function (config) {
 			{pattern: "libs/**/*.js", included: false, served: true, watch: true},
 			{pattern: "dist/**/*", included: false, served: true, watch: true},
 			{pattern: "libs/path-to-regexp.js", included: false, served: true},
+			{pattern: "libs/BezierCurve.js", included: false, served: true},
 			"tests/karma/testPaths.js",
```
TomaszLukawskiSam, line 55: I think that file`{pattern: "some-script.js", included: false, served: true},`  is also needed.

tests/karma/uicomponents.mobile.conf.js
```diff
@@ -70,7 +67,8 @@ module.exports = function (config) {
 			}
 		},
 
-		browsers: ["bigScreen"],
+		// browsers: ["bigScreen"],
+		browsers: ["ChromeHeadless"],
```
TomaszLukawskiSam: What size of screen will be after your change?

tests/karma/all.conf.js
hsiwkin: right, done

tests/karma/all.conf.js
```diff
@@ -82,16 +84,15 @@ module.exports = function (config) {
 		// available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
 		customLaunchers: {
 			bigScreen: {
-				base: "Chrome",
-				flags: ["--window-size=800,800"]
```
TomaszLukawskiSam, line 86: Why do you remove "--widndow-size", is it not supported?

TomaszLukawskiSam (changes requested):

tests/karma/all.conf.js
hsiwkin: I thought it isn't but apparently it is supported: 
https://stackoverflow.com/questions/43541925/how-to-set-the-browser-window-size-when-using-google-chrome-headless
https://peter.sh/experiments/chromium-command-line-switches/#window-size

hsiwkin: It's back

TomaszLukawskiSam (approved):
Looks good to me.

Issue #632, opened 2020-02-12
[UIComponents] Too large header title


lmslachciak:
1. UIComponents -> Graphs
2. Observe:
![image](https://user-images.githubusercontent.com/2246331/74317188-34c71580-4d7b-11ea-8be3-03dc8fc1f06c.png)
3. Expected:
normal header

lslachciak:
Issue doesn't exist in TAU 1.1

PR #634, opened 2020-02-12
Restore ui-title class creation

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/632
[Problem] Header in sample app for wearable is too large.

[Cause] Mobile profile implementation removed uititle class creation when
only header is used in example
[Solution] Restore ui-title class creation


TomaszLukawskiSam (approved):
Verified. Works and looks good.

Issue #645, opened 2020-02-14
Navigation elements --> Title --> Title bar on list


lmslachciak:
Navigation elements --> Title --> Title bar on list

PR #651, opened 2020-02-17
 Add long title on list example

lslachciak:
[Issue] #645
[Problem] no long title example
[Solution] Added long title on list example basing on the title+subtitle on list example


TomaszLukawskiSam (approved):
Looks good to me.

PR #652, opened 2020-02-18
Title with subtitle on list with 2 subitems

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/645
[Problem] No example for list with 2 sbitems
[Solution] Add example. Add css property for header.
[Remark] New css property seems similar to header-medium one but
according to requirements header title shuld be 16 px from top, so there was
a need to create new one

### DIFF
--- a/examples/wearable/UIComponents/contents/navigationelements/titlebar/title-subtitle-on-list-2lines.html
+++ b/examples/wearable/UIComponents/contents/navigationelements/titlebar/title-subtitle-on-list-2lines.html
@@ -1,0 +1,47 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta content="width=device-width, initial-scale=1, user-scalable=no" name="viewport" />
+	<title>
+		Title with subtitle on list with 2 subitems
+	</title>
+	<link href="../../../lib/tau/wearable/theme/default/tau.css" rel="stylesheet" />
+	<link href="../../../lib/tau/wearable/theme/default/tau.circle.css" rel="stylesheet" />
+	<link href="../../../css/style.css" rel="stylesheet" />
+	<link href="../../../css/style.circle.css" rel="stylesheet" />
+	<script src="../../../js/circle-helper.js">
+	</script>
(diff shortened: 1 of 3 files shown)


TomaszLukawskiSam (commented):
Spell issues in commit message:
"sbitems" -> subitems
"shuld" -> should

examples/wearable/UIComponents/contents/navigationelements/titlebar/title-subtitle-on-list-2lines.html
```diff
@@ -0,0 +1,47 @@
...
+	<link href="../../../css/style.circle.css" rel="stylesheet" />
+	<script src="../../../js/circle-helper.js">
+	</script>
+	<script src="../../../lib/tau/wearable/js/tau.js">
+	</script>
+</head>
+
+<body>
+	<div class="ui-page">
+		<header class="ui-header-on-list-with-2-subitems">
```
TomaszLukawskiSam: I'm wondering about this class name.
Seems too long. Maybe we can use name convention like for list items, eg. `title-has-2line-sub`

lslachciak: has-2line-sub can suggest 2 subtitles in title,
I was trying different abbreviation but seems no good way to shorten it

lslachciak:
spell issues fixed in commit message

examples/wearable/UIComponents/contents/navigationelements/titlebar/title-subtitle-on-list-2lines.html
TomaszLukawskiSam: As we discussed let use header medium instead of new class.

lslachciak: done

TomaszLukawskiSam (approved):
Looks good to me.

Issue #636, opened 2020-02-12
[UIComponents] Marquee List - artifact


lmslachciak:
List-->Marquee List 
Observe:
strange artifact:
![image](https://user-images.githubusercontent.com/2246331/74343044-b5e9d100-4daa-11ea-8c3b-1b229e1d3ca8.png)

lslachciak:
This is not regression

lslachciak:
looks like this example was done intentionally

lslachciak:
tau v1.0.1 - no issue
tau v1.0.20 - issue

lslachciak:
tau v1.0.10 - issue

lslachciak:
tau 1.0.5 - issue
tau 1.0.2 - no issue
tau 1.0.3 - no issue
tau 1.0.4 - no issue

lslachciak:
This is a regression after a3ca8365200f6c9215712f71f7da50c4c8af3cbb

PR #658, opened 2020-02-19
Fix ArcListView layout for multiline items

lslachciak:
[Issue] #636
[Problem] ArcListView multiline items have broken layout.
[Reason] Row layout is applied for items
[Solution] Apl solution similar to snaplist view
for multiline items use default column layout
[Remark]
Issue was introduced in a3ca836 but after further changes row layout is
effective because of 76ede39
[Test]
List > Marquee List > Obs
ArcListview > ArcListview > Obs
Control > radio > Obs
Control > Slider > Obs

Signed-off-by: Lukasz Slachciak <l.slachciak@samsung.com


TomaszLukawskiSam (approved):
Verified. Works very good.

Issue #663, opened 2020-02-20
[UIComponents] Hardware back button issue


DariuszPaziewski:
The application performs two steps back in history when you are deeper than the first level in the navigation tree and the hardware back button is used. It occurs on wearable 5.5 emulator and app with TAU 1.1.15. It does not occur on wearable 5.0.

Emulator version:
Model=Tizen5/Unified;
Build=tizen-unified_20191031.1;
Release=20191031.1;
[Build]
Type=eng;
Date=20191031_052051;
Time=05:20:51;
Variant=NONE;
ID=tizen-unified;

Steps to reproduce:
1. Make sure you use TAU 1.1.15
2. Launch the UIComponents app
3. Go to Navigation > IndexScrollBar
4. The IndexScrollbar page is displayed
5. Press the hardware back button once
6. You are in the main menu (two steps up)

Expected result:
After pressing the hardware back button the app should navigate only one step up to the Navigation page.

lslachciak:
confirmed, reproducible on 5.5 emulator, not reproducible on 5.0 emulator

lslachciak:
I prepared sample Tizen app with just content and internal links.
[UIComponents.zip](https://github.com/Samsung/TAU/files/4231404/UIComponents.zip)
It occurred that on 5.5 emulator window.history.back is called automatically when HW back key is pressed.
If I run this simple app on 5.0 no navigation back occurs after pressing window.historyback

Our sample UIComponents app for 1.1 has event listener for tizen_hwback key which calls window.history.back. That is why double back was done. Probably we need to remove this aditional listener for 5.5 from sample UIComponents app

lslachciak:
Similar situation is happening with mobile UIComponents app and mobile 5.5 simulator
[Version]
Model=Tizen5/Unified;
Build=tizen-unified_20191031.1;
Release=20191031.1;

I know how to fix this issue -
in UIComponents app , file app.js we shouldnt call window.history.back.
```
diff --git a/examples/mobile/UIComponents/js/app.js b/examples/mobile/UIComponents/js/app.js
index f27f2df..e98fe2c 100644
--- a/examples/mobile/UIComponents/js/app.js
+++ b/examples/mobile/UIComponents/js/app.js
@@ -14,8 +14,6 @@
                                        tizen.application.getCurrentApplication().exit();
                                } catch (ignore) {
                                }
-                       } else {
-                               window.history.back();
                        }
                }
        });
[...]

PR #665, opened 2020-02-21
Don't call window.history.back in sample UIComponents app

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/663
[Problem] On 5.5 emulator using HW back key cause going back two times
On 5.0 emulator there is no such issue
[Reason] Tizen 5.5 emulator has alread build-in feature for web apps to call window.history.back
No need to call it by sample app
[Solution] Remove calling window.history.back from helper file
Update samples base version to 5.5

### DIFF
--- a/examples/mobile/UIComponents/.tproject
+++ b/examples/mobile/UIComponents/.tproject
@@ -2,7 +2,7 @@
 <tproject xmlns="http://www.tizen.org/tproject">
     <platforms>
         <platform>
-            <name>wearable-5.5</name>
+            <name>mobile-5.5</name>
         </platform>
     </platforms>
     <package>
--- a/examples/mobile/UIComponents/js/app.js
+++ b/examples/mobile/UIComponents/js/app.js
@@ -14,8 +14,6 @@
 					tizen.application.getCurrentApplication().exit();
 				} catch (ignore) {
 				}
-			} else {
-				window.history.back();
 			}
 		}
 	});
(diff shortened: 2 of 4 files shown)


TomaszLukawskiSam (approved):
Ok

grzegorz-czajkowski (commented):
Will this break back button for Tizen Emulator < 5.5 ?

DohyungLim (commented):
My emulator doesn't get back button intermittently.

lslachciak:
@grzegorz-czajkowski yes, if this UIComponents app run on emulator <5.5 it will have a problem.
But please note that change is done to UIComponents app, not TAU and moreover I'm changing configuration of project
@grzegorz-czajkowski maybe we should check tizen version?

@DohyungLim which version of emulator are you using? 5.0 or 5.5? Could you show your
/etc/info.ini?

DohyungLim:
@lslachciak 
I only use 5.5 emulator.

lslachciak:
Let's wait with merging this item. We should clarify with WRT this issue

singa2000:
WRT has been updated in the Tizen 5.5.
In the latest tizen 5.5 emulator version, WRT fixed the duplicated back key issue.
Would you check the latest emulator binary (after 2020 Feb 20), please?

lslachciak:
@singa2000 thank you for information, I'll check this

lslachciak:
This should be fixed on WRT side

Issue #629, opened 2020-02-11
5. Etc - End effect


lmslachciak:
Implement end effect according to the guidelines

grzegorz-czajkowski:
Added support for using Bouncing widget with both directions Vertical and Horizontal. Exposed additional API to easy show bouncing effect instead of using drag() API.
Used this API in ArcListView - feature works fine.
Need to find a way to use drag() method for both directions or change widgets that are using Bouncing widget (like SectionChanger, Scroller, VirtualList) to use new API.

DevronB:
In addition to this, the bouncing effects at the end of lists, sections changers etc. when scrolling with drag or rotary bezel on wearable does not have the vibration feedback as it does in native.

grzegorz-czajkowski:
@DevronB 
Thanks for your comment. We will consider this as separate issue - https://github.com/Samsung/TAU/issues/667

grzegorz-czajkowski:
Main patches landed.  Follow up: https://github.com/Samsung/TAU/issues/677
Issue can be closed.

PR #670, opened 2020-02-21
Bouncing: add example to UIComponents

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/629
[Problem] Lack of Bouncing effect as specified in guide
[Solution] Add "Bouncing effect" to etc/


TomaszLukawskiSam (approved):
Verified. Looks good to me.

PR #671, opened 2020-02-24
Content: show bouncing effect on rotary

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/629
[Problem] There is no bouncing effect on rotary whereas
	on scroll it is shown. According to guide, scoll and rotary
	events should behave in the same way on content element.
[Solution] show bouncing effect on rotary event.

### DIFF
--- a/src/js/core/util/rotaryScrolling.js
+++ b/src/js/core/util/rotaryScrolling.js
@@ -39,18 +39,41 @@
 				 * Scroll step
 				 * @type {number}
 				 */
-				scrollStep = 40;
+				scrollStep = 40,
+				/**
+				 * Time after bound effect will be hidden.
+				 * @type {number}
+				 */
+				BOUNCING_TIMEOUT = 1000;
 
 			/**
 			 * Handler for rotary event
 			 * @param {Event} event Event object
 			 */
 			function rotaryDetentHandler(event) {
+				var scrollview = null,
+					maxScrollY = 0;
+
 				if (element.getAttribute("data-lock-rotary-scroll") !== "true") {
+					scrollview = ns.engine.getBinding(element.parentElement, "Scrollview");
 					if (event.detail.direction === "CW") {
 						element.scrollTop += scrollStep;
+						if (scrollview) {
+							maxScrollY = element.scrollTop + window.innerHeight;
+							scrollview.bouncingEffect.drag(undefined /* ignore horizontal effect */, -maxScrollY);
+						}
 					} else {
 						element.scrollTop -= scrollStep;
+						if (scrollview) {
+							scrollview.bouncingEffect.drag(undefined /* ignore horizontal effect */, element.scrollTop);
+						}
+					}
+
+					// Hide bouncing after timeout.
+					if (scrollview) {
+						setTimeout(function () {
+							scrollview.bouncingEffect.dragEnd();
+						}, BOUNCING_TIMEOUT);
 					}
 				}
 			}


src/js/core/util/rotaryScrolling.js
```diff
@@ -39,18 +39,41 @@
 				 * Scroll step
 				 * @type {number}
 				 */
-				scrollStep = 40;
+				scrollStep = 40,
+				/**
+				 * Time after bound effect will be hidden.
+				 * type {number}
```
TomaszLukawskiSam: `@type`

src/js/core/util/rotaryScrolling.js
```diff
@@ -39,18 +39,41 @@
...
+				var scrollview = null,
+					maxScrollY = 0;
+
 				if (element.getAttribute("data-lock-rotary-scroll") !== "true") {
+					scrollview = ns.engine.getBinding(element.parentElement, "Scrollview");
 					if (event.detail.direction === "CW") {
 						element.scrollTop += scrollStep;
+						if (scrollview) {
+							maxScrollY = element.scrollTop + window.innerHeight;
+							scrollview.bouncingEffect.drag(undefined /* ignore horizontal effect */, -maxScrollY);
```
TomaszLukawskiSam, line 63: This is little strange, this `undefined`as argument. Is this only way?

src/js/core/util/rotaryScrolling.js
```diff
@@ -39,18 +39,41 @@
 				 * Scroll step
 				 * @type {number}
 				 */
-				scrollStep = 40;
+				scrollStep = 40,
+				/**
+				 * Time after bound effect will be hidden.
+				 * type {number}
+				 */
+				bouncingTimeout = 1000;
```
TomaszLukawskiSam: I think this is const value. You can change this to 'BOUNCING_TIMEOUT'

TomaszLukawskiSam (changes requested):

src/js/core/util/rotaryScrolling.js
grzegorz-czajkowski: This can not be 0 and not maxScrollX since they will cause to show horizontal bouncing which is not we want. I think leaving it as undefined with comment make the code clear.

src/js/core/util/rotaryScrolling.js
grzegorz-czajkowski: Done.

src/js/core/util/rotaryScrolling.js
grzegorz-czajkowski: Done.

src/js/core/util/rotaryScrolling.js
```diff
@@ -39,18 +39,41 @@
...
 						element.scrollTop -= scrollStep;
+						if (scrollview) {
+							scrollview.bouncingEffect.drag(undefined /* ignore horizontal effect */, element.scrollTop);
+						}
+					}
+
+					// Hide bouncing after timeout.
+					if (scrollview) {
+						setTimeout(function () {
+							scrollview.bouncingEffect.dragEnd();
```
TomaszLukawskiSam, line 75: please use: scrollview.hideBouncingEffect()

TomaszLukawskiSam (changes requested):

src/js/core/util/rotaryScrolling.js
grzegorz-czajkowski: I added new "orientation" parameter to drag(...) method to force only vertical bouncing in case of rotary on Content. This looks quite good but reversing coordinates in Scroller (used by SecrionChanger) prevents us to use bouncing with Content (conditions are not meet).

Alternatively we could call drag() method only when end of scroll is reached (as it's done in Arclistview). This will allow calling drag(0, -maxScrollY). I leave this for future modifications.

grzegorz-czajkowski:
This needs improvements and it out of scope currently.

Issue #659, opened 2020-02-19
Wearable Section Changer - Circular:true issue


DevronB:
When using the sectionchanger on wearable with the circular:true option there is a scroll/animation issue. The issue seems to occur when scrolling to section one after the start index ('ui-section-active'). When you scroll to the end/start of the sections it correctly scrolls to the start/end in a circular motion. But when you hit the second section after the start (initial 'ui-section-active') it seems to scroll all sections end to end to return at the section you had intended to scroll to. So the behaviour is correct, but the animation is very odd and not smooth.

lslachciak:
@DevronB does the issue occurs only when rotating with bezel and no issue when swipe with finger?
I can see some strange behaviour on tau_1.0 in UIComponents sample app: Navigation-->Sectionchanger-->Circular + rotate with bezel

lslachciak:
in problematic case

translate = "translate3d(" + x + "px," + y + "px, 0)";
is always called with x set to 0 in TAU

this is because 
calculateCustomLayout returns 360
I think there is something wrong with index calculation in this function

DevronB:
Correct, I have confirmed in my case also this only occurs when using the bezel, not finger swipe.

lslachciak:
Thank you @DevronB for confirmation.

After further investigaion I can say that the problem is with
_endScroll: function () 
```
				_endScroll: function () {
					console.log("this.scrolled: " + this.scrolled + " this.scrollCanceled: " + this.scrollCanceled);
					if (!this.enabled || !this.scrolled || this.scrollCanceled) {
						return;
					}

					this._repositionSections();
					this._super();
				},

```
In case of swipe this.scrolled is always true so this._repositionSections(); is called
In case of rotarydetend this.scrolled is false so this._repositionSections(); is not called

If I ommit 
`					if (!this.enabled || !this.scrolled || this.scrollCanceled) {`
condition there is no issue with section change animation

DevronB:
Great, thank you. So is removing this code the solution, or does this.scrolled need to be set during rotarydetent event chain? Anyway when do you think a commit will be done to fix this issue?

Thank you for the fast diagnosis.

lslachciak:
@DevronB fix for this issue has just been merged.
We plan to make new TAU release this week.

DevronB:
Great, thank you. I pulled the commit and it seems to be working fine now.

lslachciak:
TAU v1.0.27 released with fix for the issue
[...]

PR #672, opened 2020-02-25
Use direct parameter to set scrolled on true in case of rotary event

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/659
[Problem] When rotating with bezel strange animation occurs in section changer
[Solution] Use direct parameter which sets scrolled value to true in setActiveSection.
This will allow to reposition sections in _endScroll in case of rotarydetent event.

### DIFF
--- a/src/js/core/widget/core/SectionChanger.js
+++ b/src/js/core/widget/core/SectionChanger.js
@@ -464,8 +464,10 @@
 
 					switch (event.type) {
 						case "swipe":
-						case "rotarydetent" :
 							this._change(event);
+							break;
+						case "rotarydetent" :
+							this._change(event, true);
 							break;
 						case "webkitTransitionEnd":
 						case "mozTransitionEnd":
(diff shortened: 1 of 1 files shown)


src/js/core/widget/core/SectionChanger.js
```diff
@@ -241,7 +241,7 @@
 						self._initLayout();
 						self._super(element);
 						self._repositionSections(true);
-						self.setActiveSection(self.activeIndex);
+						self.setActiveSection(self.activeIndex, false);
```
TomaszLukawskiSam: You were going to add "false" as default value to 'direct' but you set default value for duration. Please check source code for `setActiveSection`

TomaszLukawskiSam: You can set default values for method like: 
```
				setActiveSection: function (index, duration, direct) {
					var position = this.sectionPositions[index],
						scrollbarDuration = duration || 0,
						oldActiveIndex = this.activeIndex,
						newX = 0,
						newY = 0;

					direct = !!direct;
```

TomaszLukawskiSam (changes requested):

src/js/core/widget/core/SectionChanger.js
lslachciak: done

src/js/core/widget/core/SectionChanger.js
```diff
@@ -502,18 +504,21 @@
...
 				 * the duration parameter must be in milliseconds.
-				 * @param {number} [direct=false]
+				 * @param {number} [direct=false] Wheter section is set once directly (e.g. bezel)
+				 *  or with touch events
 				 * @member ns.widget.core.SectionChanger
 				 */
 				setActiveSection: function (index, duration, direct) {
 					var position = this.sectionPositions[index],
-						scrollbarDuration = duration,
+						scrollbarDuration = (duration = (duration || 0)),
```
TomaszLukawskiSam: Understanding this expression requires too much effort for me.

I prefer simple (safe) solution like:
```
var position = this.sectionPositions[index],
      scrollbarDuroation,
      oldActiveIndex = this.activeIndex,
      newX = 0,
      newY = 0;

duration = duration || 0;
scrollbarDuration = duration;
...
```

lslachciak: done

TomaszLukawskiSam (approved):
Looks good to me.

Issue #628, opened 2020-02-11
 1. Navigation elements - Go-to-top-button - content


lmslachciak:
1. Navigation elements - Go-to-top-button - content

lslachciak:
depends on https://github.com/Samsung/TAU/issues/625

grzegorz-czajkowski:
Two commits merged. Issue can be closed.

Issue #627, opened 2020-02-11
 1. Navigation elements - Go-to-top-button - list


lmslachciak:
1. Navigation elements - Go-to-top-button - list

lslachciak:
depends on https://github.com/Samsung/TAU/issues/625

grzegorz-czajkowski:
Patch merged. Notice bug https://github.com/Samsung/TAU/issues/725 which will be handled separately.

PR #681, opened 2020-02-27
Page: add Go-to-top button

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/628
[Problem] No Go-to-top button in content example
[Solution] Add Go-to-top button to Page widget according to
   the guideline and implement it for Content.
[Remarks] On lists, the button does not work since it will be
   handled in https://github.com/Samsung/TAU/issues/627.
    

src/css/profile/wearable/changeable/common/button.less
```diff
@@ -355,3 +355,10 @@ button.ui-btn, input.ui-btn {
 		margin-top: 0;
 	}
 }
+
+.ui-button-go-to-top {
+	width: 52 * @unit_base;
+	height: 52 * @unit_base;
+	background-image: url("./images/GoToTopButton/go_to_top_dark.png");
```
TomaszLukawskiSam: I think we shouldn't create additional directory like 'GoToTopButton' to store images. This makes additional issues when images will be updated.

TomaszLukawskiSam (changes requested):

grzegorz-czajkowski:
As we discussed, I am changing the approach to add the button to the page (instead of scrollview). This will allow us to handle click on the button for arclistview.

src/css/profile/wearable/changeable/common/button.less
grzegorz-czajkowski: We are having images in dirs https://github.com/Samsung/TAU/tree/tau_1.0/src/css/profile/wearable/changeable/theme-changeable/images

grzegorz-czajkowski:
Added commit to enable go top button by Page option.

src/css/profile/wearable/changeable/common/button.less
TomaszLukawskiSam: I thought about not creating additional subdirectories `GoToTopButton`

TomaszLukawskiSam (approved):
Looks good to me.

PR #687, opened 2020-03-03
Reposition sections when setActiveSection API is called by app creator

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/659
[Problem] When calling setActiveSection API directly in app, in case of swipe strange animation occurs
[Reason] Normally _repositionSections method is called when widget is initialized.
When ui-section-active not set all values will be initialized with default ones.
When developers calls setActiveSection directly method _repositionSections is not called

[Solution] Call _repositionSections inside setActiveSection

### DIFF
--- a/src/js/core/widget/core/SectionChanger.js
+++ b/src/js/core/widget/core/SectionChanger.js
@@ -241,7 +241,7 @@
 						self._initLayout();
 						self._super(element);
 						self._repositionSections(true);
-						self.setActiveSection(self.activeIndex);
+						self.setActiveSection(self.activeIndex, 0, false, false);
 
 						// set correct options values.
 						if (!options.animate) {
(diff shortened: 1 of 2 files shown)


src/js/core/widget/core/SectionChanger.js
```diff
@@ -510,7 +510,7 @@
 				 *  or with touch events
 				 * @member ns.widget.core.SectionChanger
 				 */
-				setActiveSection: function (index, duration, direct) {
+				setActiveSection: function (index, duration, direct, reposition) {
```
TomaszLukawskiSam, line 514: Please add js doc for `reposition`.

TomaszLukawskiSam (changes requested):

src/js/core/widget/core/SectionChanger.js
lslachciak: done

TomaszLukawskiSam (approved):

HunseopJeong (approved):
LGTM

PR #752, opened 2020-03-11
List (mobile): Update the UX of the List

HunseopJeong:
[Issue] N/A
[Problem] Need to update UX for List
[Solution] Apply 5-1 and 6-4

### DIFF
--- a/examples/mobile/UIComponents/components/list/basiclist.html
+++ b/examples/mobile/UIComponents/components/list/basiclist.html
@@ -18,17 +18,39 @@
 		</div>
 		<div class="ui-content">
 			<ul class="ui-listview ui-content-area">
-				<li>
-					Main text 01
+				<li class="ui-li-divider">
+					<div class="ui-li-text">
+						<span class="ui-li-text-title">
+							Main text 01
+						</span>
+						<span class="ui-li-text-sub ui-li-text-value">
+							Value text
+						</span>
+					</div>
+				</li>
+				<li class="ui-li-divider">
+					<div class="ui-li-text">
+						<span class="ui-li-text-title">
+							Main text 02
+						</span>
+					</div>
+				</li>
+				<li class="ui-li-divider">
+					<div class="ui-li-text">
+						<span class="ui-li-text-title">
+							Main text 03
+						</span>
+						<span class="ui-li-text-sub">
+							This is the second line for descriptive phrases. Chunking background color is #fcfcfc.
+						</span>
+					</div>
 				</li>
 				<li>
-					Main text 02
-				</li>
-				<li>
-					Main text 03
-				</li>
-				<li>
-					Main Text 04
+					<div class="ui-li-text">
+						<span class="ui-li-text-title">
+							Main Text 04
+						</span>
+					</div>
 				</li>
 			</ul>
 			<button data-icon="add" data-style="fab"></button>
--- a/src/css/profile/mobile/common/oneui-common.less
+++ b/src/css/profile/mobile/common/oneui-common.less
@@ -16,6 +16,7 @@
     --background-color: @background-color;
     --background-area-color: @background-area-color;
 
+    --text-color: @text-color;
     --tab-text-color: @tab-text-color;
     --tab-text-color-dim: fade(@tab-text-color, @tab-text-opacity);
     --popup-background: @popup-background-color;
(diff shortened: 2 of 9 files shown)


HunseopJeong:
Snapshot
image

examples/mobile/UIComponents/components/list/basiclist.html
```diff
@@ -19,16 +19,41 @@ <h1>
...
+					<div class="ui-li-text">
+						<span class="ui-li-text-title">
+							Main text 01
+						</span>
+						<span class="ui-li-text-sub ui-li-text-value">
+							Value text
+						</span>
+					</div>
 				</li>
+				<div class="ui-li-divider"></div>
```
TomaszLukawskiSam: 1. This is wrong HTML syntax, `<ul>` shouldn't contains `div` elements.
2. I think we don't need "ui-li-divider", please use just "border-bottom" css property.

TomaszLukawskiSam (changes requested):

examples/mobile/UIComponents/components/list/basiclist.html
TomaszLukawskiSam: We can use `ui-li-divider` between lists eg. 
```
<ul>
   <li></li>
</ul>
<div class="ui-li-divider"></div>
<ul>
   <li></li>
</ul>
```

TomaszLukawskiSam (changes requested):

examples/mobile/UIComponents/components/list/basiclist.html
TomaszLukawskiSam: One correction. We cannot use border-bottom because the line is not on whole width of list item. We have to use pseudo element `::before` or `::after`.

HunseopJeong: Okay, I removed the "div.ui-li-divider" and move it into the "li.ui-li-divider".

src/css/profile/mobile/themes/themes.less
```diff
@@ -40,7 +40,6 @@ body.ui-theme-dark {
     // eg. --primary-color: @primary-color;
     @import (multiple) "../common/oneui-common.less";
 
-    --text-color: @text-color;
```
TomaszLukawskiSam, line 43: You cannot remove this line because you broke colors in other themes.
image

TomaszLukawskiSam (changes requested):

src/css/profile/mobile/themes/themes.less
HunseopJeong: I miss to move it to the "onui-common.less".
I fixed it.

src/css/profile/mobile/common/oneui-common.less
```diff
@@ -16,6 +16,7 @@
     --background-color: @background-color;
     --background-area-color: @background-area-color;
 
+    --text-color: @text-color;
```
TomaszLukawskiSam, line 19: this variable has already been added in `themes.less` line: 43

src/css/profile/mobile/themes/light.variables.less
```diff
@@ -26,6 +26,8 @@
     @feature-card-underline-color: #eaeaea;
     @feature-card-subtitle-color: #43a6b0;
     @flicker-pointer-color: #ABD7EB;
+
+    @text-color: #252525;
```
TomaszLukawskiSam, line 30: this property is already defined in `base.less`

TomaszLukawskiSam (changes requested):

src/css/profile/mobile/themes/light.variables.less
HunseopJeong: Question)
Why we define colors for the light theme to the base.less?
Is it default value for the light.theme?
Can I define it to the light.variable.less?

HunseopJeong: Is there any reasons to define the "--text-color" to the :root ?
Can I remove the "@text-color" and --text-color in base.less?

TomaszLukawskiSam: Good question. 
The `base.less` was supposed to contain common values.
`light.less` and `dark.less` was supposed to contain differences.
You can move `@text-color` to light theme if think that it will be better but try don't duplicate variables.

HunseopJeong: IMO, values that do not change depending on the theme shoud be in base.less.
Such as "primary-color".
So I think moving to "oneui-common.less" looks good because "text-color" changes depding on the theme.

HunseopJeong: There are a lot of values that are defined incorrectly, I will try to re-arrange them later.

TomaszLukawskiSam (approved):
Looks good to me.

singa2000 (approved):
It looks good~

Issue #737, opened 2020-03-10
Controllers -> On/Off switch


lmslachciak:

lslachciak:
Controllers -> On/Off switch

TomaszLukawskiSam:
I verified colors and sizes. In my opinion all is correct. I made little fix for example in UIComponents.

TomaszLukawskiSam:
Patch has already been merged.

lslachciak:
example should be updated when master switch on/off is done
(example in guideline contains master switch too)

lslachciak:
works well, thank you

PR #756, opened 2020-03-12
On-Off Switch: fix for example in UIComponents

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/737
[Problem] Verify implementation according to guideline
[Solution] Example in UIComponents has deprecated list style


lslachciak (approved):
works well

PR #758, opened 2020-03-13
List (mobile): Update the UX of the List

HunseopJeong:
[Issue] N/A
[Problem] Need to update the UX for List
[Solution] Apply 5-2, 6-6, and 6-7

### DIFF
--- a/src/css/profile/mobile/common/core.less
+++ b/src/css/profile/mobile/common/core.less
@@ -148,11 +148,20 @@
 
 	.ui-content-area {
 		background-color: var(--background-area-color);
-		mask-border: url(images/0_Round_corner/round.svg) 77 round;
+		mask-border: url(images/0_Round_corner/round.svg) 77;
 		mask-border-width: 26 * @px_base;
 		border-radius: 26 * @px_base;
 		border: 1 * @px_base solid var(--content-area-line-color);
 		box-sizing: border-box;
+	}
+
+	.ui-content-subheader {
+		color: var(--text-secondary-color);
+		.font(medium);
+		font-size: 14 * @sp_base;
+		padding-bottom: 5 * @px_base;
+		margin-top: -4.25 * @px_base;
+		margin-left: 24 * @px_base;
 	}
 }
 
--- a/src/css/profile/mobile/common/listview.less
+++ b/src/css/profile/mobile/common/listview.less
@@ -42,7 +42,7 @@ tau-expandable {
 	transform: translate3d(0, 0, 0);
 
 	// End of List (Bottom spacing)
-	&.ui-content-area:last-of-type {
+	&.ui-content-area {
 		margin-bottom: 16 * @px_base;
 
 		// Fab button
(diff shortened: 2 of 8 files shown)


HunseopJeong:
Snapshot
image

src/css/profile/mobile/common/core.less
```diff
@@ -148,12 +148,19 @@
 
 	.ui-content-area {
 		background-color: var(--background-area-color);
-		mask-border: url(images/0_Round_corner/round.svg) 77 round;
-		mask-border-width: 26 * @px_base;
```
HunseopJeong, line 152: I removed mask border because the wrong line appeared in specific resolution.
image
Is there any problem to remove the mask border ? I think it can be covered by radius border.. :(

TomaszLukawskiSam: You have to check svg files for corners eg.  `images/0_Round_corner/bottom_left.svg`. If you sure that the rounding effect will be the same you can use `border-radius`. At the moment the effect seems the same.

src/css/profile/mobile/common/listview.less
```diff
@@ -61,12 +61,64 @@ tau-expandable {
...
+			align-items: center;
+
+			&:empty {
+				margin-top: 16 * @px_base;
+			}
+
+			.ui-li-text-subheader {
+				display: inline-flex;
+				color: var(--text-secondary-color);
+				font-family: Roboto-Medium;
```
TomaszLukawskiSam: For family font we have `less` method `.font(medium)`, `.font(regular)`, `.font(light)`.

TomaszLukawskiSam: You can check this method in tizen.less

src/css/profile/mobile/themes/dark.variables.less
```diff
@@ -43,3 +43,4 @@
 
     @divider-color: #d4d4d4;
     @divider-opacity: 15%;
+    @subheader-divider-color: #fafafa;
```
TomaszLukawskiSam: this is the same as: @color-whie, so
`@subheader-divider-color: @color-white`

src/css/profile/mobile/common/listview.less
HunseopJeong: Good Tip! I will check it.

src/css/profile/mobile/common/core.less
HunseopJeong: I removed 'round' from the mask-border and fixed issue.

src/css/profile/mobile/themes/dark.variables.less
HunseopJeong: I moved the @color-white to the "base.less". and change it use the @color-white.

TomaszLukawskiSam (approved):
Great! Thank you.

singa2000 (approved):
It's good~

PR #760, opened 2020-03-16
(mobile): Update the UIComponent App

HunseopJeong:
[Issue] N/A
[Problem] Listview appears unexpectedly
[Solution] Apply changed listview UX

### DIFF
--- a/examples/mobile/UIComponents/index.html
+++ b/examples/mobile/UIComponents/index.html
@@ -20,68 +20,68 @@
 			</h1>
 		</header>
 		<div class="ui-content">
-			<ul class="ui-listview ui-colored-list">
-				<li>
+			<ul class="ui-listview ui-content-area">
+				<li class="ui-li-anchor">
 					<a href="components/appbar/index.html">
 						App Bar
 					</a>
 				</li>
-				<li>
+				<li class="ui-li-anchor">
 					<a href="components/controllers/index.html">
 						Controllers
 					</a>
 				</li>
-				<li>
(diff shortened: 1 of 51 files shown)


HunseopJeong:
Snapshot
image

examples/mobile/UIComponents/index.html
```diff
@@ -20,68 +20,68 @@ <h1>
 			</h1>
 		</header>
 		<div class="ui-content">
-			<ul class="ui-listview ui-colored-list">
-				<li>
+			<ul class="ui-listview ui-content-area">
+				<li class="ui-li-anchor">
```
singa2000, line 24: I just have a question.
If it use with divider, is it right  as     li class="ui-li-divider ui-li-anchor" ?

HunseopJeong: Yes, It is right. But I don't think the 'ui-li-divider' is used by default.

lslachciak: I think this is sth what we need to clarify with UX if divider is used by default or not

lslachciak (approved):

examples/mobile/UIComponents/index.html
singa2000: In the UX , 
List basically has a divider.
But in case of the list used in Dialog, it basically doesn't have a divider.
User have a choice by "ui-li-divider" and now there is no problem for it.

singa2000 (approved):
It's good.

Issue #753, opened 2020-03-11
Appbar --> Navigate up buttom -> no action


grzegorz-czajkowski:
User is not navigated backwards after click on back button in appbar examples.

Observations:
 * data-rel="back" is set
 * the proper behavior can be seen after replacing button element to link, for example <a href="" ...></a>
 * changes on router side are probably needed

grzegorz-czajkowski:
Patch merged. Back button works. This change caused wrong width of ui-appbar-left-icons-container. Reported here https://github.com/Samsung/TAU/issues/764

lslachciak:
thank you 
 works well

PR #763, opened 2020-03-16
Appbar (Mobile): Change back button to link

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/753
[Problem] User is not navigated backwards after click on back button in appbar examples.
[Solution] Replace button to link (<a></a>) since this element is tracked TAU router.

[Remarks] This change causes wrong width (8px) of ui-appbar-left-icons-container but it
	should have 32px. Only right margin (8px) of link element is taken into account.
	It looks like icon size (24x24) is not properly added.


TomaszLukawskiSam (commented):
Please create issue for wrong width of "ui-appbar-left-icons-container"

grzegorz-czajkowski:
@TomaszLukawskiSam Done

TomaszLukawskiSam (approved):
Looks good to me.

PR #766, opened 2020-03-17
List (mobile): Update List examples to apply changed UX

HunseopJeong:
[Issue] N/A
[Problem] After changing the UX, existing sreen of exmaples is weird.
[Solution] Apply the changed UX to examples(on-off,selectall, reorder, and expandable)

### DIFF
--- a/examples/mobile/UIComponents/components/list/reorder.html
+++ b/examples/mobile/UIComponents/components/list/reorder.html
@@ -21,189 +21,186 @@
 		</div>
 		<div class="ui-content">
 			<ul class="ui-listview ui-content-area" id="drag-list">
-				<li class="ui-li-flex">
-					<span class="ui-li-area ui-li-area-a">
-						<span class="ui-li-text">
-							<span>
-								Primary text
-							</span>
-						</span>
-					</span>
-				</li>
-				<li class="ui-li-flex">
-					<span class="ui-li-area ui-li-area-a">
-						<span class="ui-li-text">
-							<span>
-								Primary text
-							</span>
-							<span class="ui-li-text-sub-2">
-								Sub Text 2
-							</span>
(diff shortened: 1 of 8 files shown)


HunseopJeong:
On-off elements
image
Selectable
image
Reorder
image
image

TomaszLukawskiSam:
@lslachciak @HunseopJeong 
This PR looks good to me but we have to consider which current listview examples we need in UIComponents app and which examples should be removed or updated.

HunseopJeong:
Yes. We need to consider which examples should be updated or deleted.
There are so many duplicated examples and outside of scope of the UX guide.

examples/mobile/UIComponents/components/list/reorder.html
```diff
@@ -229,16 +224,14 @@ <h1>
...
-							<span>
-								Primary text
-							</span>
+				<li>
+					<div class="ui-li-text">
+						<span class="ui-li-text-title">
+							Multi primary text
 						</span>
-						<span class="ui-li-text-2">
+						<span class="ui-li-text-sub ui-li-text-value">
```
DohyungLim, line 234: Do you use sub text instead of 2?
If so, how do you handle cases of 3 or higher?

DohyungLim (commented):
I think that code other than 1 comment looks fine.

examples/mobile/UIComponents/components/list/reorder.html
HunseopJeong: I changed original usages.
We can place the text to the correct position by using the 'ui-li-text-title' and 'ui-li-text-sub'.

singa2000 (approved):
Now it  looks  good.
If it's needed, other samples will be update in later.

Issue #741, opened 2020-03-10
Controllers -> Slider --> Continuous Slider


lmslachciak:

lslachciak:
Looks like after 849eefb542a0b8fe7ab2e25f8323886a58fa2624 Slider was broken

TomaszLukawskiSam:
border-image: url(images/3_Controllers/tw_progress_bar_n.9.png) 5 fill;
    border-image-width: 3px;
    border-image-repeat: round;

lslachciak:
Thank you @TomaszLukawskiSam I was trying this on Friday but without 
border-image-repeat: round;
option. I'll check.

lslachciak:
Currently facing issue with setting opacity not for whole element but only border-image
CSS doesnt allow this directly
https://stackoverflow.com/questions/45085415/border-image-with-opacity
If I set opactiy for .ui-slider-bar it will also reflect .ui-slider-value

lslachciak:
All current items completed. When new png/svg provided I'll update them

PR #775, opened 2020-03-18
Slider example outside of list

lslachciak:
[Problem] As a result of 849eefb sliders can not be on list anymore
[Solution] Put sliders outside of list.


lslachciak:
Other slider changes/style updates will be in separate PRs under https://github.com/Samsung/TAU/issues/741

HunseopJeong (approved):
I have a plan to support the slider in the list but not now.

grzegorz-czajkowski (approved):

PR #793, opened 2020-03-20
Update the Flat button example

singa2000:
[Issue] N/A
[Problem] The Flat button example is weird.
[Solution] Update the Flat button with list.


lslachciak (approved):
verified, it looks much better now

PR #794, opened 2020-03-20
Dropdown list (Mobile): Min width

grzegorz-czajkowski:
Two commits:
 * UIComponents (Mobile): add Dropdown list (Min width)
 * Dropdown list (Mobile): apply alignment according to guide

### DIFF
--- a/src/css/profile/mobile/common/dropdownmenu.less
+++ b/src/css/profile/mobile/common/dropdownmenu.less
@@ -179,12 +179,9 @@
 	overflow: hidden;
 	z-index: 1201;
 	min-width: 168px;
-	&.ui-dropdownmenu-options-top {
+	&.ui-dropdownmenu-options-vertical-margins {
 		margin-top: 6 * @px_base;
-	}
-	&.ui-dropdownmenu-options-bottom {
 		margin-bottom: 6 * @px_base;
-
 	}
 	&.ui-dropdownmenu-active {
 		visibility: visible;
@@ -193,6 +190,11 @@
 	&.ui-dropdownmenu-options-opened {
 		border-top: 1 * @px_base solid @color_dropdownmenu_list_group_stroke;
 		border-bottom: 1 * @px_base solid @color_dropdownmenu_list_group_stroke;
+	}
+	// top, bottom classes just describes direction in which menu is animated.
+	&.ui-dropdownmenu-options-top {
+	}
+	&.ui-dropdownmenu-options-bottom {
 	}
 	&.ui-dropdownmenu-options-opening.ui-dropdownmenu-options-top .ui-dropdownmenu-options {
 		.animation(open-to-top 300ms);
--- a/src/js/profile/mobile/widget/DropdownMenu.js
+++ b/src/js/profile/mobile/widget/DropdownMenu.js
@@ -179,6 +179,9 @@
 						elContent: null,
 						elDefaultOption: null
 					};
+
+					self._horizontalPosition = null;
+
 					/**
 					 * @property {Object} options Object with default options
 					 * @property {boolean} [options.nativeMenu=true] Sets the DropdownMenu widget as native/custom type.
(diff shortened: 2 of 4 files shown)


src/js/profile/mobile/widget/DropdownMenu.js
```diff
@@ -1010,17 +1011,19 @@
...
 				parseInt(stylesOfSelectedOptionAfter.width, 10) : wrapperMinWidth;
 
 				height = optionHeight;
+
 				// This part decides the location and direction of option list.
-				offsetLeft = self._offset.left;
+				offsetLeft = self._horizontalDirection === "right" ? window.screen.width - width : 0;
 				optionStyle = "left: " + offsetLeft + "px; ";
+
 				if (options.inline === true) {
```
DohyungLim, line 1032: Type is boolean, do I need "true"? If it can be omitted, nothing is concise.

DohyungLim (commented):
1 comment need to check.

src/js/profile/mobile/widget/DropdownMenu.js
grzegorz-czajkowski: Are you refering to this line?
`if (options.inline === true)`
I am not changing this and could be refactored separately.

grzegorz-czajkowski:
Solved most issues related to alignment position
 - added top and bottom margins if dropdown is only opened from non list item element (for example appabar)
 - dropdow list is shown of left or right most position based on click position
 - fixed alignment on list (avoid addition 12px from top, added from js code).
Created additional issue https://github.com/Samsung/TAU/issues/808 where we need to take into consideration list position to always show dropdown menu within list.

src/css/profile/mobile/common/dropdownmenu.less
```diff
@@ -194,6 +191,11 @@
 		border-top: 1 * @px_base solid @color_dropdownmenu_list_group_stroke;
 		border-bottom: 1 * @px_base solid @color_dropdownmenu_list_group_stroke;
 	}
+	// top, bottom classes just describes direction in which menu is shown.
```
TomaszLukawskiSam: you don't need define empty styles.

src/js/profile/mobile/widget/DropdownMenu.js
TomaszLukawskiSam: I agree with Grzegorz. This is out of scope this PR.

src/js/profile/mobile/widget/DropdownMenu.js
```diff
@@ -578,30 +587,28 @@
 			};
 
 			/**
-			 * Return offset of element
-			 * @method getOffsetOfElement
+			 * Return top offset of element
+			 * @method getTopOffsetOfElement
```
TomaszLukawskiSam, line 591: You modified description but method name remains old.

src/js/profile/mobile/widget/DropdownMenu.js
```diff
@@ -1073,12 +1086,14 @@
 			 * @protected
```
TomaszLukawskiSam, line 1087: @param {event} [clickEvent=null] click event

TomaszLukawskiSam (changes requested):

src/css/profile/mobile/common/dropdownmenu.less
grzegorz-czajkowski: Right but they are still needed.
Those classes (top or bottom) are applied by DropdownMenu.js and are used by css to animate showing menu in specified direction, for example:
```
        &.ui-dropdownmenu-options-opening.ui-dropdownmenu-options-top .ui-dropdownmenu-options {
                .animation(open-to-top 300ms);
        }
```

src/js/profile/mobile/widget/DropdownMenu.js
grzegorz-czajkowski: Done.

src/js/profile/mobile/widget/DropdownMenu.js
grzegorz-czajkowski: Done.

TomaszLukawskiSam (approved):
Works good.

PR #796, opened 2020-03-20
Listview: Left / right margin issue

TomaszLukawskiSam:
[Issue] N/A
[Problem] Margins on right and left sides make a confusion
[Solution]
 At the moment, because of list divider, all list elements like texts, button, icons
 have to have special padding 4px. The list divider can be changed for
 removing this special padding

### DIFF
--- a/src/css/profile/mobile/common/listview.less
+++ b/src/css/profile/mobile/common/listview.less
@@ -57,7 +57,7 @@ tau-expandable {
 		display: flex;
 		position: relative;
 		box-sizing: content-box;
-		margin: 0 20 * @px_base;
+		margin: 0 23 * @px_base;
 		overflow: visible;
 		text-align: left;
 		font-size: @font_size_list_main_text;
@@ -68,7 +68,7 @@ tau-expandable {
 
 		// Subheader
 		&.ui-li-subheader {
-			margin-left: 20 * @px_base;
+			margin-left: 23 * @px_base;
 			flex-direction: row;
 			align-items: center;
 
@@ -139,7 +139,7 @@ tau-expandable {
 		.ui-li-text {
 			display: inline-flex;
 			flex-direction: column;
-			padding: 14 * @px_base 4 * @px_base;
+			padding: 14 * @px_base 0;
 
 			.ui-li-text-title {
 				font-size: 18 * @sp_base;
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam:
@HunseopJeong 
Hi, I tryied to positioning button on listview and I saw that because of listdivider all elements require special padding on left or right side. I think we can optimize this by this patch. Can you look this PR? Thank you.

HunseopJeong:
@TomaszLukawskiSam,
I think you need to check the 5-1 Divider list.
The divider has left and right 20 margins. And many components has different left margins started from divider. So I applied the different margin to the each components. What is the Button example you think is difficult to position? Once implemented, I'll check it again according to the guide.

TomaszLukawskiSam:
> @TomaszLukawskiSam,
> I think you need to check the 5-1 Divider list.
> The divider has left and right 20 margins. And many components has different left margins started from divider. So I applied the different margin to the each components. What is the Button example you think is difficult to position? Once implemented, I'll check it again according to the guide.

I think that only divider has 20 margin. Most of items on list have 24px margin (6-2). I didn't see the components set relative to the divider in guideline. Maybe I missed them. I do not insist, we can stay with the current state.

HunseopJeong:
@TomaszLukawskiSam,
You can check it on icon page (6-5).
Icon has different margins. So I changed the divider width and left when divider has icon.
```
			&.ui-li-has-icon::after {
				width: calc(~"100%" - 52 * @px_base);
				left: 52 * @px_base;
			}
```

TomaszLukawskiSam:
> @TomaszLukawskiSam,
> You can check it on icon page (6-5).
> Icon has different margins. So I changed the divider width and left when divider has icon.
> 
> ```
> 			&.ui-li-has-icon::after {
> 				width: calc(~"100%" - 52 * @px_base);
> 				left: 52 * @px_base;
> 			}
> ```

But in this case you have additional padding between icon and text "Main text".

HunseopJeong:
It has 12 margins between icon and main text without additional work.
But I checked it there is some issue with left margin when it has icon... :( I will check it.

TomaszLukawskiSam:
> It has 12 margins between icon and main text without additional work.
> But I checked it there is some issue with left margin when it has icon... :( I will check it.

Sorry my mistake.

TomaszLukawskiSam:
> > It has 12 margins between icon and main text without additional work.
> > But I checked it there is some issue with left margin when it has icon... :( I will check it.
> 
> Sorry my mistake.

You are right. You did smaller margin because of this padding.
You see, this is the confusion, because these values is not defined strict as in guideline but are related to others elements.

HunseopJeong:
> 
> 
> > It has 12 margins between icon and main text without additional work.
> > But I checked it there is some issue with left margin when it has icon... :( I will check it.
> 
> Sorry my mistake.

No. My mistake :( I will check it.
Anyway I think this PR is not proper now. If I need to change it, I will request it to you!

TomaszLukawskiSam:
ok.

Issue #782, opened 2020-03-19
Listview - Examples are need to be rearranged 


HunseopJeong:
There are many examples for the list, but there are many duplicates, so it seems necessary to organize them.

Current Index for the list
**Stacked List**
- Basic List
- List Styles
- Reorder List
- Normal List
- Selectable List
- Expandable
- Expandable from JS
- List with On-Off elements
- Virtual List

**Grid View**
- Grid View
- Grid View with Label (outside)
- Grid View with Label (inside)
- Grid View with Checkbox
- Grid View with Badge
- Grid View with Move
- Grid View within Popup

HunseopJeong:
I think we can remove the List Styles, and Normal List or combine two into own.
@lslachciak, @TomaszLukawskiSam,
How about your opinion?
If there's more to be removed, you can tell me.

lslachciak:
Personally I would prefer to follow guideline and have just submenus like:
- List Width
- List contents area
- End of list
- Basic list
- Icon + text
- Subheader
- Expandalble list
- multiple selection

So according to your list I would leave only 
- Basic List
- Selection List
- Expandable List
And move rest somewhere to backup
The other list examples according to GUI will need to be provided in future.

TomaszLukawskiSam:
I think we don't need backup. We have repository and complete version of UIComponents on branch 1.1

HunseopJeong:
I reogranize the list example following structrue.
- List Width
- List Content Area
- End of list
-- Bottom spacing
-- Bottom spacing with FAB
- Basic List
- Icon + Text List
- Subheader
-- Subheader (Text)
-- Subheader (No text)
-- Subheader (Without chunking)
- Expandable List
- Multiple Selection List

lslachciak:
Good idea

HunseopJeong:
PR merged

PR #798, opened 2020-03-23
Update the menu for the List

HunseopJeong:
[Issue] #782
[Problem] There was existing ooutdate examples
[Solution] Remove outdate examples and update it followed by UX guide


HunseopJeong:
Multiple Selection List is empty because I didn't implement it yet.
I will implement it after merging this PR.

DohyungLim (approved):
I think this PR no problem.

LGTM.

examples/mobile/UIComponents/components/list/basiclist.html
```diff
@@ -45,100 +51,57 @@ <h1>
...
-					</span>
-				</li>
-				<li class="ui-li-subheader">
-					<span class="ui-li-text-subheader">
-						Subheader
-					</span>
-				</li>
 				<li class="ui-li-divider">
-					<span class="ui-li-text">
+					<div class="ui-li-text">
```
singa2000, line 102: I just have a question how to use it.
To use the "ui-li-text-title" and ""ui-li-text-sub", do I have to use "ui-li-text"?
Can I use "ui-li-text-title" directly without "ui-li-text"??

HunseopJeong: 'ui-li-text' is the container for texts because title and sub text need to place it vertically (To apply flex-direction: column).
So I define the 'ui-li-text' and include the title and sub. And the container has the padding values, so we need to include it to apply the padding to texts.

HunseopJeong:
Changed PR to apply kebab case.

singa2000 (approved):
It looks good.
Thanks

Issue #802, opened 2020-03-23
File naming rules in UI Components app


HunseopJeong:
In UI components app, various naming rules are applied.
Ex) In appbar, used the camel case "actionButton"
      In controlls, used the kebab case "checkbox-image-thumbnail.html"
      In graphs, used the snake case "bar_graph.html"
I think we need to unify in one way.
How about your opinions?
Which is best?

TomaszLukawskiSam:
In my opinion, most are in format "kebab". I prefer this rule.
`checkbox-image-thumbnail.html`

HunseopJeong:
Any other opinions? If not, I will create the PR to unify it with kebab case?
After that PR, please follow the rule.

HunseopJeong:
@lslachciak, @singa2000, @DohyungLim 
Any other opinions?

singa2000:
I also prefer to kebab.
It seems that Vue and Angular follows kebab like  https://angular.io/guide/styleguide.
It looks in general HTML use the kebab.

HunseopJeong:
Okay. I will prepare the PR to apply kebab case.
I hope you will check this rule as you review.

TomaszLukawskiSam:
@HunseopJeong 
Can we close the issue?

HunseopJeong:
Issue was fixed.

PR #803, opened 2020-03-23
Rename files with kebab case style

HunseopJeong:
[Issue] #802
[Problem] Several naming rules are used irregularly
[Solution] Rename files with kebab case style


HunseopJeong:
It is hard to judge which of 'searchbar' and 'search-bar' is correct.
So I just change the camel and the snake cases to the kebab case.

TomaszLukawskiSam (approved):
Looks good to me.

DohyungLim (approved):
LGTM.

DohyungLim (commented):
you need to check conflict branch.

lslachciak (approved):
for wearable profile I think we will need unificaiton too

HunseopJeong:
> 
> 
> for wearable profile I think we will need unificaiton too

Good Think :D We would be good to apply same rules!

singa2000 (approved):
It's good.
I will follow the coding rule.

PR #804, opened 2020-03-23
[mobile] Style active and inactive slider bar

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/741
[Problem] Styling not in accordance to guideline
[Solution] Use peudo element for enabled slider. Use border-image CSS property.

### DIFF
--- a/src/css/profile/mobile/common/slider.less
+++ b/src/css/profile/mobile/common/slider.less
@@ -6,18 +6,27 @@
 	flex-direction: column;
 	justify-content: space-around;
 
-	.ui-slider-bar {
-		background-color: var(--slider-bg-color);
-		height: 3 * @px_base;
-		border-radius: 1.5 * @px_base;
-		overflow: hidden;
-		transition: height 250ms, border-radius 250ms;
-		transition-timing-function: cubic-bezier(0.33, 0, 0.2, 1);
+	.ui-slider-bar::before {
+		content: '';
+		left: 0;
+		top: 0;
+		width: 100%;
+		height: 100%;
+		border-image: url(images/3_Controllers/tw_progress_bar_n.9.png) 3 fill;
+		border-image-width: 1 * @px_base;
+		opacity: 0.3;
+		position: absolute;
+	}
 
+	.ui-slider-bar {
+		position: relative;
+		height: 3 * @px_base;
+		overflow: hidden;
 
 		.ui-slider-value {
 			height: 100%;
-			background-color: var(--slider-handler-color);
+			border-image: url(images/3_Controllers/tw_progress_bar_n.9.png) 3 fill;
+			border-image-width: 1 * @px_base;
 		}
 	}
 
@@ -36,6 +45,7 @@
 		transition: transform cubic-bezier(0.33, 0, 0.2, 1) 300ms,
 					left cubic-bezier(0.17, 0.2, 0.2, 1) 200ms;
 		transform-origin: 0 50% 0;
+		z-index: 9;
 	}
 
 	&.ui-slider-active {
@@ -46,12 +56,14 @@
 	}
 
 	&.ui-disabled {
-		opacity: 1;
 		.ui-slider-bar {
-			background-color: var(--slider-bg-disabled-color);
+			border-image: url(images/3_Controllers/tw_progress_bar_d.9.png) 3 fill;
+			border-image-width: 1 * @px_base;
+			opacity: 0.15;
 
 			.ui-slider-value {
-				background-color: var(--slider-handler-disabled-color);
+				border-image: url(images/3_Controllers/tw_progress_bar_d.9.png) 3 fill;
+				border-image-width: 1 * @px_base;
 			}
 		}
 


lslachciak:
current status:
image

lslachciak:
known issues: 
1. disabled slider handler should not be transprent
2. No round corners of slider bar

lslachciak:
Issue with z handler resolved. Please review
I would like to face rounded corners issue separately (in separate PR)

Current status
image

src/css/profile/mobile/common/slider.less
```diff
@@ -6,18 +6,30 @@
...
+		content: '';
+		left: 0;
+		top: 0;
+		width: 100%;
+		height: 100%;
+		border-image: url(images/3_Controllers/tw_progress_bar_n.9.png) 5 fill;
+		opacity: 30%;
+		border-radius: 1.5 * @px_base;
+		position: absolute;
+
```
HunseopJeong: Remove empty line.

TomaszLukawskiSam: You need two things:
```
border-image-width: 3px;
border-image-repeat: round;
```

src/css/profile/mobile/common/slider.less
```diff
@@ -46,17 +58,18 @@
...
 
 			.ui-slider-value {
-				background-color: var(--slider-handler-disabled-color);
+				border-image: url(images/3_Controllers/tw_progress_bar_d.9.png) 5 fill;
 			}
 		}
 
 		.ui-slider-handler {
 			background-color: var(--slider-handler-disabled-color);
+			z-index: 9;
```
HunseopJeong: Question) Is there any reasons to use the '9'?

src/css/profile/mobile/common/slider.less
TomaszLukawskiSam: > You need two things:
> 
> ```
> border-image-width: 3px;
> border-image-repeat: round;
> ```

I checked again and I got a better result for:
` border-image-width: 1px;` without `  border-image-repeat: round;`

lslachciak: done

src/css/profile/mobile/common/slider.less
lslachciak: not exactly, I just needed to give it higher riority than other elements

lslachciak:
image

lslachciak:
Updated PR.
image
After making some experiments slicing with 3 looks the best.
As you can see on the above screenshot there is a shadow, but this is because original png file which has shadow too
https://github.com/Samsung/TAU/blob/tau_1.2/src/css/profile/mobile/theme-changeable/images/3_Controllers/tw_progress_bar_n.9.png

Also on Tizen 5.5 emulator example doesn't looks so nice like on Android device, probably due to the missing support for some CSS properites
image

HunseopJeong (approved):
LGTM

src/css/profile/mobile/common/slider.less
```diff
@@ -6,18 +6,27 @@
...
 	justify-content: space-around;
 
+	.ui-slider-bar::before {
+		content: '';
+		left: 0;
+		top: 0;
+		width: 100%;
+		height: 100%;
+		border-image: url(images/3_Controllers/tw_progress_bar_n.9.png) 3 fill;
+		border-image-width: @px_base;
```
TomaszLukawskiSam: this entry seems like trick... can we write more clear: `1 * @px_base;`

DohyungLim (commented):
It's a simple question.
Is the slider made in the form of a widget? I wondered if the location would come out well if I added it to a Dialog or a table.

singa2000 (commented):
It is necessary to check whether the 9-patch works properly not only in Android but also in Tizen device. Because as I known, the 9-patch usually use for Andorid device.
So we need to check it also properly working in the WebApp and Tizen device.

lslachciak:
I'll check reason of slider not working properly on Tizen 5.5

lslachciak:
> 
> 
> It's a simple question.
> Is the slider made in the form of a widget? I wondered if the location would come out well if I added it to a Dialog or a table.

Slider is the same widget as other TAU widgets. But we test it in the specific circumstances as in guideline. E.g. Currently, after recent list changes slider is not supported on the list, see https://github.com/Samsung/TAU/pull/775 for more details

lslachciak:
The problem which was with Tizen 5.5 was an opacity CSS property
I provided values with % but the are supported since Chrome 78
https://developer.mozilla.org/en-US/docs/Web/CSS/opacity
Tizen 5.5 seems to have Chrome 47 version
image

src/css/profile/mobile/common/slider.less
lslachciak: done

lslachciak:
uploaded new with changes
- fix accoring to the @TomaszLukawskiSam comment about px_base
- correct support for Tizen 5.5 (opacity % changed to decimal points)
now it works fine also on Tizen 5.5
image

TomaszLukawskiSam (approved):
Looks good to me.

Issue #764, opened 2020-03-17
Appbar --> Navigate up button -> wrong width of ui-appbar-left-icons-container


grzegorz-czajkowski:
After https://github.com/Samsung/TAU/pull/763, there is wrong width (8px) of ui-appbar-left-icons-container but it should have 32px. Only right margin (8px) of link element is taken into account.
It looks like icon size (24x24) is not properly added.

lslachciak:
@TomaszLukawskiSam @grzegorz-czajkowski can we close this issue?

grzegorz-czajkowski:
I will verify this.

grzegorz-czajkowski:
Works fine. Thank you.

PR #806, opened 2020-03-23
Appbar: fix for width of back button in appbar

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/764
[Problem] Wrong width of ui-appbar-left-icons-container
[Solution]
 - Back button in icon container has not width because of
  display property


HunseopJeong (approved):
LGTM

Issue #809, opened 2020-03-24
Margin in list element


HunseopJeong:
Currently, li element has the default left and right margin with 20px.
But many components have their own magin values such as the checkbox (18px), radio (18px), and icon (12px). 
So I think it would be good to remove the default margin of the li elmeent.
It can help to apply margins of components.

HunseopJeong:
Issue was resolved.

PR #786, opened 2020-03-19
Apply click style for slider handler

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/741
[Problem] Slider handler style not up-to-date for click
[Solution] Apply guide style

Signed-off-by: Lukasz Slachciak <l.slachciak@samsung.com>

PR #811, opened 2020-03-24
Change to calculate the margin in list

HunseopJeong:
[Issue] #809
[Problem] Many components has their own magins, so it is hard to apply margin for each component
[Solution] Remove the default margin of the li element and apply the margin for each component

### DIFF
--- a/src/css/profile/mobile/common/listview.less
+++ b/src/css/profile/mobile/common/listview.less
@@ -57,7 +57,6 @@ tau-expandable {
 		display: flex;
 		position: relative;
 		box-sizing: content-box;
-		margin: 0 20 * @px_base;
 		overflow: visible;
 		text-align: left;
 		font-size: @font_size_list_main_text;
@@ -91,7 +90,7 @@ tau-expandable {
 			&::after {
 				content: "";
 				display: inline-flex;
-				width: 100%;
+				width: calc(~"100%" - 20 * @px_base);
 				background-color: var(--subheader-divider-color);
 				mask-image: url(images/6_Lists/tw_list_subheader_divider.svg);
 				mask-size: 4.5 * @px_base 1.5 * @px_base;
@@ -99,13 +98,15 @@ tau-expandable {
 				mask-position: center;
 				height: 1.5 * @px_base;
 				align-items: center;
+				margin-right: 20 * @px_base;
 			}
 		}
 
 		&.ui-li-anchor {
 			display: flex;
 			flex-direction: column;
-			padding: 14 * @px_base 4 * @px_base;
+			padding: 14 * @px_base 0;
+			margin: 0 24 * @px_base;
 		}
 
 		// Icon
@@ -115,14 +116,11 @@ tau-expandable {
 			display: inline-flex;
 			position: relative;
 			align-self: center;
+			margin: 0 12 * @px_base;
 
 			&.ui-li-icon-small, &.ui-li-icon-small:after {
-				width: 25px !important;
-				height: 25px !important;
-			}
-
-			+ .ui-li-text {
-				margin-left: 8px;
+				width: 25 * @px_base !important;
+				height: 25 * @px_base !important;
 			}
 
 			&::after, * {
@@ -139,7 +137,8 @@ tau-expandable {
 		.ui-li-text {
 			display: inline-flex;
 			flex-direction: column;
-			padding: 14 * @px_base 4 * @px_base;
+			padding: 14 * @px_base 0;
+			margin-right: 24 * @px_base;
 
 			.ui-li-text-title {
 				font-size: 18 * @sp_base;
@@ -160,13 +159,18 @@ tau-expandable {
 				color: var(--primary-color);
 			}
 		}
+
+		& > .ui-li-text:first-child {
+			margin-left: 24 * @px_base;
+		}
 		
 		// Divider
 		&.ui-li-divider {
 			&::after {
 				content: "";
 				position: absolute;
-				width: 100%;
+				width: calc(~"100%" - 40 * @px_base);
+				left: 20 * @px_base;
 				bottom: 0;
 				height: 1 * @px_base;
 				background-color: var(--divider-color);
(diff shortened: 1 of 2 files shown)


HunseopJeong:
@TomaszLukawskiSam,
It seems to affect the button as well, so I modified it.
Is there any place I should check?

src/css/profile/mobile/common/listview.less
```diff
@@ -188,7 +191,7 @@ tau-expandable {
 			&::after {
 				content: "";
 				position: absolute;
-				width: 1 * @px_base;
+				width: @px_base;
```
TomaszLukawskiSam, line 195: For me this format makes confusion, eg.
`padding: 0 3 * @px_base @px_base 0`

HunseopJeong: You mean 1 * @px_base looks better than @px_base?

TomaszLukawskiSam: I mean, it's compatible with the general format:
`0` or `1px` ... `10px` 
`0` or `1 * @px_base` ... `10 * @px_base`

HunseopJeong: Yes,, But, Do we use the @px_base for the scability?
Anyway,, I don't understand what I do.. :(

src/css/profile/mobile/common/listview.less
```diff
@@ -105,7 +104,8 @@ tau-expandable {
 		&.ui-li-anchor {
 			display: flex;
 			flex-direction: column;
-			padding: 14 * @px_base 4 * @px_base;
+			padding: 14 * @px_base 0;
+			margin: 0 24 * @px_base;
```
TomaszLukawskiSam, line 109: I think in current implementation we have to take account width of border line (1px)

src/css/profile/mobile/common/listview.less
lslachciak: we use @px_base to allow proper dp-->CSS PX convertion
currently 1 dp = 1 CSS PX
but there may be devices with different scale in future

src/css/profile/mobile/common/listview.less
TomaszLukawskiSam: I'm wondering how reduce this border line. Is possible to use border-image style like in slider widget.
https://github.com/Samsung/TAU/pull/786
Similar to:
```
border-image: url(images/3_Controllers/tw_progress_bar_n.9.png) 5 fill;
border-image-width: 3px;
border-image-repeat: round;
```
but of course different image (with transparency) and different parameters.

HunseopJeong: Do we consider the border width? Why?

HunseopJeong: In 6-2, the text has 24 margin on the left and right.
Do we consider the border line at this case?

TomaszLukawskiSam: There is outline stroke. In current implementation this line width are added to margin and in fact elements on list are positioned 25px.

HunseopJeong: You mean the border width of the 'ui-content-area'?
I see.. Okay,, I will try to find how to resolve that issue.

HunseopJeong: I think we can fix border issue with outline.
I will fix it in next PR.

HunseopJeong:
@TomaszLukawskiSam, @lslachciak , @singa2000,
Could you review this PR?

src/css/profile/mobile/common/listview.less
```diff
@@ -160,22 +158,27 @@ tau-expandable {
...
 				bottom: 0;
 				height: 1 * @px_base;
 				background-color: var(--divider-color);
 				opacity: var(--divider-opacity);
 			}
 
 			&.ui-li-has-icon::after {
-				width: calc(~"100%" - 52 * @px_base);
-				left: 52 * @px_base;
+				width: calc(~"100%" - 84 * @px_base);
```
DohyungLim, line 181: just question.
Is responsive applied so -84px is maintained to the specification?

src/css/profile/mobile/common/listview.less
```diff
@@ -139,7 +136,8 @@ tau-expandable {
 		.ui-li-text {
 			display: inline-flex;
 			flex-direction: column;
-			padding: 14 * @px_base 4 * @px_base;
+			padding: 14 * @px_base 0;
```
DohyungLim, line 140: Can I omit the trailing 0?

src/css/profile/mobile/common/listview.less
```diff
@@ -115,14 +115,11 @@ tau-expandable {
...
 
 			&.ui-li-icon-small, &.ui-li-icon-small:after {
-				width: 25px !important;
-				height: 25px !important;
-			}
-
-			+ .ui-li-text {
-				margin-left: 8px;
+				width: 25 * @px_base !important;
+				height: 25 * @px_base!important;
```
DohyungLim: need to 1 space :)

DohyungLim (commented):
3 comment check.

src/css/profile/mobile/common/listview.less
HunseopJeong: Yes. It works correctly because it is calculated base on the 100%.

src/css/profile/mobile/common/listview.less
HunseopJeong: I don't understand your suggestion,, why omit the trailing '0'?

TomaszLukawskiSam: This is correct for me.

src/css/profile/mobile/common/listview.less
HunseopJeong: Done

TomaszLukawskiSam:
I verified this PR with guideline. I found one issue with subheader
image

and one knowing issue with 1px border of content area.

HunseopJeong:
> 
> 
> I verified this PR with guideline. I found one issue with subheader
> image
> 
image
I fixed it with margin-right: 20px

> and one knowing issue with 1px border of content area.

I will check it in next PR.

TomaszLukawskiSam (approved):
Looks good to me.

singa2000 (approved):
It looks good~

PR #812, opened 2020-03-24
Fonts: fix for font definition

TomaszLukawskiSam:
[Issue] N/A
[Problem] Medium font is too bold
[Solution] Font definition has been fixed

### DIFF
--- a/src/css/profile/mobile/common/tizen.less
+++ b/src/css/profile/mobile/common/tizen.less
@@ -12,14 +12,13 @@
 @unit_base: (360 / 720) * @px_base;
 
 .font(light) {
-	font-family: Roboto-Light, sans-serif-light;
+	font-family: Roboto-Light;
 }
 .font(regular) {
-	font-family: Roboto-Regular, sec-roboto-light;
+	font-family: Roboto-Regular;
 }
 .font(medium) {
-	font-family: Roboto-Medium, sec-roboto-light;
-	font-weight: bold;
+	font-family: Roboto-Medium;
 }
 
 //*****************************


HunseopJeong (approved):
I think it too bold.

src/css/profile/mobile/fonts.less
```diff
@@ -1,16 +1,18 @@
 @font-face {
-    font-family: Roboto-Light;
+    font-family: sans-serif-light;
     src: url(fonts/Roboto-Light.ttf);
 }
 
 @font-face {
-    font-family: Roboto-Regular;
+    font-family: sec-roboto-light;
```
lslachciak: according to the MDN font-family
Specifies a name that will be used as the font face value for font properties.
if we have two the same font-family names how we can differentiate them?

src/css/profile/mobile/common/tizen.less
```diff
@@ -12,13 +12,14 @@
...
+	font-family: sans-serif-light;
 }
 .font(regular) {
-	font-family: Roboto-Regular, sec-roboto-light;
+	font-family: sec-roboto-light;
+	font-weight: normal;
 }
 .font(medium) {
-	font-family: Roboto-Medium, sec-roboto-light;
+	font-family: sec-roboto-light;
```
lslachciak: I though that sec-robot-light is alternative font if the Roboto-Medium is not available

lslachciak (changes requested):
I have doubt regarding this PR.
I think that we should clarify with UX first:
1. Does Roboto Medium should have bold style? If yes, why in guides when Roboto MEdium is used font is not bold?
2. What should be priority
sec
roboto light over Roboto Regular.ttf when exist in system?
roboto light

HunseopJeong:
It used the origin medium in subheader
image

It is the subheader image at UX guide (6-6)
image

It seems to be different in my eyes... :(
I knew that we followed the UX guide but why does it look different...

singa2000:
@HunseopJeong 
OK.
As I see,  Roboto Medium should have bold style and the he GUI Guide screenshot also use the bold style.
But it different look as you told.
I will check it.

singa2000:
I checked about the font.

1. When we use "Roboto Medium", we should not  put the "font-weight: bold"
    Would you just remove the "font-weight", please?

    .font(medium) {
	font-family: Roboto-Medium, sec-roboto-light;
	_font-weight: bold;_
}

2. About the font-family like "sec-xx" , we don't need to consider it.

lslachciak:
@TomaszLukawskiSam I agree with @singa2000 lets remove 
font-weight: bold;
and
"sec-xx

TomaszLukawskiSam:
> @TomaszLukawskiSam I agree with @singa2000 lets remove
> font-weight: bold;
> and
> "sec-xx

I updated PR.

HunseopJeong (approved):
LGTM

lslachciak (approved):
Buttons now looks much better, verified

PR #822, opened 2020-03-25
Update Dialogs Sample

DohyungLim:
- Change the text and name of the sample.
- Button and Controller list will be changed later.
- 3 Stacked button is not implemented.

### DIFF
--- a/examples/mobile/UIComponents/components/dialogs/dialog.html
+++ b/examples/mobile/UIComponents/components/dialogs/dialog.html
@@ -20,67 +20,67 @@
 			<ul class="ui-listview ui-content-area">
 				<li class="ui-li-anchor">
 					<a data-position-to="window" data-rel="popup" href="#2-btn-popup">
-						2 Button Title Dialog
+						2 Button Basic Dialog
 					</a>
 				</li>
 				<li class="ui-li-anchor">
-					<a data-position-to="window" data-rel="popup" href="#1-btn-2level-popup">
-						1 Button Title 2level description Dialog
+					<a data-position-to="window" data-rel="popup" href="#2level-desc-dialog">
+						Title + 2level description Dialog
 					</a>
 				</li>
 				<li class="ui-li-anchor">
-					<a data-position-to="window" data-rel="popup" href="#1-btn-checkbox-popup">
-						1 Button Title Checkbox Dialog
+					<a data-position-to="window" data-rel="popup" href="#checkbox-dialog">
+						Checkbox Dialog
 					</a>
 				</li>
 				<li class="ui-li-anchor">
-					<a data-position-to="window" data-rel="popup" href="#2-btn-notitle-popup">
-						2 Button No Title Dialog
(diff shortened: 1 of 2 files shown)


examples/mobile/UIComponents/components/dialogs/dialog.html
```diff
@@ -20,67 +20,67 @@ <h1>
...
 				</li>
 				<li class="ui-li-anchor">
-					<a data-position-to="window" data-rel="popup" href="#0-btn-title-popup">
-						0 Button Title Dialog
+					<a data-position-to="window" data-rel="popup" href="#no-action-dialog">
+						No Action Dialog
 					</a>
 				</li>
 				<li class="ui-li-anchor">
 					<a data-inline="true" data-position-to="window" data-rel="popup" href="#listview-popup">
```
HunseopJeong: How about changing the #listview-popup to the #listview-dialog to sync with title.

src/css/profile/mobile/common/popup.less
```diff
@@ -57,6 +57,7 @@
 		overflow: hidden;
 		text-overflow: ellipsis;
 		white-space: nowrap;
+		font-family: Roboto-Medium;
```
HunseopJeong: Use the ".font(medium)"

examples/mobile/UIComponents/components/dialogs/dialog.html
DohyungLim: I see. fixed.

done.

src/css/profile/mobile/common/popup.less
DohyungLim: I modify your comment.

done.

src/css/profile/mobile/common/popup.less
```diff
@@ -144,13 +149,7 @@
...
 
 	.ui-listview {
-		border-top-width: 1 * @px_base;
-		border-top-color: #e6e6e6;
-		border-top-style: solid;
-		border-bottom-width: 1 * @px_base;
-		border-bottom-color: #e6e6e6;
-		border-bottom-style: solid;
-		margin-left: -16 * @px_base;
+		margin-left: -23 * @px_base;
```
HunseopJeong: Is there any reason to set the margin left with - 23px?

DohyungLim: Dialog itself has padding on the left side. 
The listview inserted inside also has a margin on the left, so it looks skewed to the right.
so, to adjust the spacing, type a value of -23px.

HunseopJeong: Is the 24px of the dialog padding?
But why do you move to the left with -23px?

DohyungLim: When the header font was roboto-regular, 23px was an appropriate value.
But for roboto-medium, 24px is appropriate.
I will change it.

HunseopJeong (approved):
LGTM

Issue #821, opened 2020-03-25
Controllers --> Checkbox on list --> Ripple effec


lmslachciak:

lslachciak:
Reassigning PR to @TomaszLukawskiSam he wants to finish this item

TomaszLukawskiSam:
Second patch has been merged. I think that we can close the issue.

lslachciak:
works great

PR #824, opened 2020-03-25
Add active class support for checkbox

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/821
[Problem] Unable to recognize touch start and end for riple effect in CSS
[Solution] Add active class support in JS

### DIFF
--- a/src/js/core/widget/core/Checkbox.js
+++ b/src/js/core/widget/core/Checkbox.js
@@ -67,7 +67,8 @@
 				},
 				classes = {
 					checkbox: "ui-checkbox",
-					focus: "ui-checkbox-focus"
+					focus: "ui-checkbox-focus",
+					active: "ui-checkbox-active"
 				},
 				prototype = new BaseWidget();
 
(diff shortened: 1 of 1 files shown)


lslachciak:
this is first step in order to provide ripple effect for checkbox

src/js/core/widget/core/Checkbox.js
```diff
@@ -167,6 +168,20 @@
 				}
 			};
 
+			prototype._onTouchStart = function () {
+				var self = this,
```
TomaszLukawskiSam: you don't need these local variables, enough:
`this.element.classList.add(classes.active);`

Creation of local variables make sense when we use a object property more that once.
This rule has reason in js file minification. The local variables will be minified to one char in opposition to object property which cannot be minified.

src/js/core/widget/core/Checkbox.js
```diff
@@ -167,6 +168,20 @@
...
+					element = self.element;
+
+				element.classList.add(classes.active);
+			};
+
+			prototype._onTouchEnd = function () {
+				var self = this,
```
TomaszLukawskiSam: you don't need these local variables, enough:
`this.element.classList.add(classes.active);`

TomaszLukawskiSam (changes requested):

src/js/core/widget/core/Checkbox.js
lslachciak: I was basing on the _onBlur method which is above

lslachciak: done

src/js/core/widget/core/Checkbox.js
lslachciak: done

lslachciak:
Fixed accordng to @TomaszLukawskiSam comments, alse added documentation for methods

TomaszLukawskiSam (approved):
Looks good to me.

Issue #825, opened 2020-03-25
ui-content-area border width


HunseopJeong:
'ui-content-area' has the border width with 1px.
In this case, it may affect other components and may not be placed in a desired location.

HunseopJeong:
PR merged

PR #827, opened 2020-03-25
Reduce border width to fix the margin issue

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/825
[Problem] ui-content-area has border width with 1px. It can affect other components.
[Solution] Use the outline instead of the border.

### DIFF
--- a/src/css/profile/mobile/common/core.less
+++ b/src/css/profile/mobile/common/core.less
@@ -151,8 +151,8 @@
 		mask-border: url(images/0_Round_corner/round.svg) 77;
 		mask-border-width: 26 * @px_base;
 		border-radius: 26 * @px_base;
-		border: 1 * @px_base solid var(--content-area-line-color);
 		box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


src/css/profile/mobile/common/core.less
```diff
@@ -151,8 +151,9 @@
 		mask-border: url(images/0_Round_corner/round.svg) 77;
 		mask-border-width: 26 * @px_base;
 		border-radius: 26 * @px_base;
-		border: 1 * @px_base solid var(--content-area-line-color);
+		outline: 1 * @px_base auto var(--content-area-line-color);
```
TomaszLukawskiSam: The patch works very good.
I don't know why we need outline? Can you show me use case?

HunseopJeong: @TomaszLukawskiSam,
Yes. you are right. The outline is not needed. I will remove it.

DohyungLim (approved):
LGTM.

PR #828, opened 2020-03-25
Dropdown list (Mobile): "Max width" example and implementation

grzegorz-czajkowski:
Two commits:
 *  UIComponents (Mobile): add "Max width" dropdown example
 *  Dropdown list (Mobile): fit to viewport

### DIFF
--- a/src/css/profile/mobile/common/dropdownmenu.less
+++ b/src/css/profile/mobile/common/dropdownmenu.less
@@ -220,7 +220,7 @@
 		overflow-y: auto;
 		background-color: var(--dropdown-menu-options-background);
 		border-radius: 26 * @px_base;
-		box-shadow: inset 0 -3 * @px_base 0 0 @dropdown-list-box-shadow-color;
+		box-shadow: 0 0 3 * @px_base 0 @dropdown-list-box-shadow-color;
 		border: var(--dropdown-menu-options-border);
 		&:focus {
 			outline: none;
--- a/src/css/profile/mobile/common/dropdownmenu.less
+++ b/src/css/profile/mobile/common/dropdownmenu.less
@@ -1,3 +1,5 @@
+@dropdown-options-margin: 6 * @px_base;
+
 .ui-dropdownmenu-overlay {
 	opacity: 0;
 	position: absolute;
@@ -179,9 +181,10 @@
 	overflow: hidden;
 	z-index: 1201;
 	min-width: 168px;
+	max-width: 100vw;
 	&.ui-dropdownmenu-options-vertical-margins {
-		margin-top: 6 * @px_base;
-		margin-bottom: 6 * @px_base;
+		margin-top: @dropdown-options-margin;
+		margin-bottom: @dropdown-options-margin;
 	}
 	&.ui-dropdownmenu-active {
 		visibility: visible;
@@ -213,6 +216,8 @@
 		list-style: none;
 		padding: 0;
 		margin: 0;
+		max-height: calc(~"100vh - "2 * @dropdown-options-margin);
+		overflow-y: auto;
 		background-color: var(--dropdown-menu-options-background);
 		border-radius: 26 * @px_base;
 		box-shadow: inset 0 -3 * @px_base 0 0 @dropdown-list-box-shadow-color;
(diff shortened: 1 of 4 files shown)


src/css/profile/mobile/base.less
```diff
@@ -76,6 +76,8 @@
 @slider-value-color: @slider-base-color;
 @slider-handler-color: @slider-base-color;
 @content-area-line-color: #d6d6d6;
+
+@dropdown-list-margin: 6 * @px_base;
 @dropdown-list-box-shadow-color: fade(@_black, 35%);
```
TomaszLukawskiSam: Try to find and remove this style:
image

TomaszLukawskiSam (changes requested):

src/css/profile/mobile/common/dropdownmenu.less
```diff
@@ -179,9 +179,10 @@
 	overflow: hidden;
 	z-index: 1201;
 	min-width: 168px;
+	max-width: 100vw;
 	&.ui-dropdownmenu-options-vertical-margins {
-		margin-top: 6 * @px_base;
-		margin-bottom: 6 * @px_base;
+		margin-top: @dropdown-list-margin;
```
TomaszLukawskiSam: do you need this variable for something else? If not, you can remove it.

src/css/profile/mobile/common/dropdownmenu.less
```diff
@@ -213,6 +214,8 @@
 		list-style: none;
 		padding: 0;
 		margin: 0;
+		max-height: calc(~"100vh - "2 * @dropdown-list-margin);
```
TomaszLukawskiSam: I think that better use simple `2 * 6 * @px_base` instead of additional variables.

TomaszLukawskiSam (changes requested):

src/css/profile/mobile/base.less
grzegorz-czajkowski: Done.

src/css/profile/mobile/common/dropdownmenu.less
grzegorz-czajkowski: Yes, I am using it to set container max-height.

src/css/profile/mobile/common/dropdownmenu.less
grzegorz-czajkowski: Wouldn't be better to have variable for this? So we will have one place to change margin and after this container position and height are updated as well?

src/css/profile/mobile/theme-changeable/theme.color.less
```diff
@@ -265,7 +265,6 @@
 //						DropdownMenu
 //***************************************************************************
 @color_dropdownmenu_list_group_stroke: B0733L1; // #[color] list stroke line color
-@color_dropdownmenu_list_item_bg_press: B042; // #[color] list item background press color
```
TomaszLukawskiSam, line 268: I think I did mistake in comment.
You can remove this line and replace this variable in:
image
by `var(--ripple-color)`

src/css/profile/mobile/common/dropdownmenu.less
```diff
@@ -213,6 +214,8 @@
...
+		overflow-y: auto;
 		background-color: var(--dropdown-menu-options-background);
 		border-radius: 26 * @px_base;
 		box-shadow: inset 0 -3 * @px_base 0 0 @dropdown-list-box-shadow-color;
```
TomaszLukawskiSam: My comment is out of scope this change but I think that the `box-shadow` is different as in guideline. `0px 0px 3px 0px rgba(0,0,0,35)`

src/css/profile/mobile/theme-changeable/theme.color.less
TomaszLukawskiSam: But I don't sure if the press effect is on dropdown menu items.

src/css/profile/mobile/common/dropdownmenu.less
TomaszLukawskiSam: I understand you point of view but you need search this value across the project now. In my opinion, if you want to leave this variable you should define it in this file eg. on beginning.

grzegorz-czajkowski: Ok, moved to the beginning of this file.

src/css/profile/mobile/common/dropdownmenu.less
grzegorz-czajkowski: css box-shadow has different kind of two first parameters (not including inset option). They are (x,y) positions of shadow box not size of border in px. That's why we can not apply them one by one.
I assume the guide intention of
`Box-shadow: 0px 0px 3px 0px rgba(0,0,0,35)`
is to add shadow on each edge, probably starting clockwise.

src/css/profile/mobile/theme-changeable/theme.color.less
grzegorz-czajkowski: Yes, @lslachciak noticed that ripple-effect is missing. However, proposed change is not sufficient to have ripple-effect working - it does not disappear. There is a separate issue https://github.com/Samsung/TAU/issues/832 for this and let's take care of this separately.

src/css/profile/mobile/common/dropdownmenu.less
TomaszLukawskiSam: https://developer.mozilla.org/pl/docs/Web/CSS/box-shadow

grzegorz-czajkowski: Ok will try this.

grzegorz-czajkowski: I applied this change, only rounded borders are having shadow.. Added separate commit fixing this. Thank you.

TomaszLukawskiSam (approved):
Ok. Looks good to me.

PR #844, opened 2020-03-27
Upate CONTRIBUTING.md with merging policy

lslachciak:
[Issue] N/A
[Problem] No detailed merging policy
[Solution] Add description which merging option use to merge PR so the history is straight

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,3 +61,5 @@ TomaszLukawskiSam
 
 * PR can be merged by PR owner or TAU reviewer if at least one TAU Reviewer approved and verified it.
 * TAU Reviewers can not approve their own PR.
+* When merging **standard PR use "Rebase and merge"** option in order to have straight git history
+* When merging changes between tau_* branches use "Create a merge commit" option in order to allow to synchronization between branches.


korneliakobiela (approved):
LGTM :+1:

TomaszLukawskiSam (approved):
ok

grzegorz-czajkowski (approved):

CONTRIBUTING.md
```diff
@@ -61,3 +61,5 @@ TomaszLukawskiSam
 
 * PR can be merged by PR owner or TAU reviewer if at least one TAU Reviewer approved and verified it.
 * TAU Reviewers can not approve their own PR.
+* When merging **standard PR use "Rebase and merge"** option in order to have straight git history
```
singa2000, line 64: How about the "Squash and merge" option?
Sometimes the merged PR remained  "Merge pull request #000  from Samsung/xxx".
I think it doesn't good look.

CONTRIBUTING.md
```diff
@@ -61,3 +61,5 @@ TomaszLukawskiSam
...
+* When merging changes between tau_* branches use "Create a merge commit" option in order to allow to synchronization between branches.
```
singa2000, line 65: What's the "Create a merge commit" option?
Could you explain it in more detail?

CONTRIBUTING.md
lslachciak: "Rebase and merge" option will all to avoid such "merge" commit.
It puts the commit on the top of history so history is straight and there is no "merge" in history (similar to git rebase)
In case of "squash and merge" few commits which are part of PR are squashed into one and merged. This is not good for clear history of changes. Sometimes we may want to split some change into few commits intentionally. "squash and merge" will destroy our intentions

CONTRIBUTING.md
lslachciak: This is standard merge, similar to git merge command. It merges histories of two branches.
It is good when synchronizing branches with each other (e.g. tau_1.0 <--> tau_1.1 <--> tau_1.2)
When we have common history between branches, git is able to automatically put changes from one branch to the second one with git merge command.
If we decide to use "rebase" option instead of merge in the next synchronization we will have problems because rebased commits common to both branches will not be recognized.

I think the core differences are nicely described here:
https://www.atlassian.com/git/tutorials/merging-vs-rebasing

CONTRIBUTING.md
DohyungLim: @lslachciak 
It is not good for us to have a merge or pull history. 
and developers always rebase themselves before merging commits. 
this is because the code is not split and merged properly. 
If it is necessary to 'rebase and merge', it proceeds, but I think it is unnecessary because git proceeds with rebase. (This is my subjective thought.)

DohyungLim (commented):
1 comment.

lslachciak:
_> It is not good for us to have a merge or pull history._
@DohyungLim I totally agree with you.
That is why I propose "Rebase and merge" option in github in README to guarantee straight history.
I dont' know better alternatives. Do you know any?
"Squash and merge" - will squash commits
"Create a merge commit" - will include merge history

Github offers only three below
image

singa2000 (approved):
OK. I agree  by default option is "Rebase and merge"

Issue #847, opened 2020-03-27
App Bar > Add "All" checkbox


TomaszLukawskiSam:
This checkbox joins with listview and toggle on checkboxes in listview.

TomaszLukawskiSam:
Patch has already been merged. The issue can be close.

PR #848, opened 2020-03-27
AppBar: Checkbox All

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/847
[Problem] Checkbox "All" is not implemented
[Solution]
 - added css styles for checkbox "All"
    
image

image

### DIFF
--- a/examples/mobile/UIComponents/components/list/multiple-selection-list.html
+++ b/examples/mobile/UIComponents/components/list/multiple-selection-list.html
@@ -13,8 +13,9 @@
 	<div class="ui-page" id="checkbox-list-selected-demo">
 		<header>
 			<div class="ui-appbar-left-icons-container">
-				<label>All
-					<input class="ui-checkbox-all" type="checkbox"/>
+				<label class="ui-label-select-all">
+					<input type="checkbox"/>
+					All
 				</label>
 			</div>
 			<div class="ui-appbar-title-container">
--- a/src/css/profile/mobile/common/appbar.less
+++ b/src/css/profile/mobile/common/appbar.less
@@ -93,15 +93,17 @@
 					height: 48 * @px_base;
 				}
 			}
+			&.ui-btn-icon-back {
+				margin-left: 20 * @px_base;
+				margin-right: 12 * @px_base;
+			}
 		}
 	}
 
 
 	.ui-appbar-controls-container {
 		flex: 0 0 56 * @px_base;
-		width: calc(~"100% - "20 * @px_base);
-
-		margin-left: 20 * @px_base;
+		width: 100%;
 
 		display: flex;
 		flex-direction: row;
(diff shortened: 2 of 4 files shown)


examples/mobile/UIComponents/components/list/multiple-selection-list.html
```diff
@@ -13,8 +13,9 @@
 	<div class="ui-page" id="checkbox-list-selected-demo">
 		<header>
 			<div class="ui-appbar-left-icons-container">
-				<label>All
-					<input class="ui-checkbox-all" type="checkbox"/>
+				<label class="ui-checkbox-all">
```
HunseopJeong: How about changing the name of the 'ui-checkbox-all' to the 'ui-checkbox-label'?
Wouldn't it be good to consider other names as well?

src/css/profile/mobile/common/appbar.less
```diff
@@ -219,4 +220,20 @@
...
+	.ui-checkbox-all {
+		font-size: 12 * @sp_base;
+		width: 32 * @px_base;
+		display: flex;
+		flex-direction: column;
+		align-items: center;
+		padding-top: 10 * @px_base;
+		line-height: 10 * @px_base;
+		margin-left: 18 * @px_base;
+		margin-right: 18 * @px_base;
```
HunseopJeong, line 233: If there is no margin for the top and bottom, it would be good to use the 'margin: 0 18*px_base'?

TomaszLukawskiSam: I am against setting redundant styles. TAU selectors are very strong and many times are hard to overwritten. If I don't have to, I try not to create them.

examples/mobile/UIComponents/components/list/multiple-selection-list.html
TomaszLukawskiSam: I didn't think too much about it. This can be named also "ui-checkbox-label-all". This is rather specific component for only one purpose.

src/css/profile/mobile/common/appbar.less
HunseopJeong: I agree it!

examples/mobile/UIComponents/components/list/multiple-selection-list.html
HunseopJeong: In my opinion, it is better to use it in a broader name than a specific purpose in terms.
How about your opinions @lslachciak, @singa2000 ?

lslachciak: for `<li>` element we have classes like `ui-li-has-checkbox ui-li-divider`
so to be common in style we should have "`label`" in name
so I think this should be named according to the `ui-label-* `schema
e.g.
`ui-label-select-all`

HunseopJeong: Okay,, `ui-label-select-all` seems to be more reasonable.

lslachciak (approved):
Select All works properly on the List --> Multiple Selectin List example, 
no integration with bottom bar but this will be a separate issue.
Please fix travis error
image

TomaszLukawskiSam:
@lslachciak 
CI issue has been fixed

lslachciak (approved):

Issue #836, opened 2020-03-26
View controls -> Dropdown list -> Two lines item


grzegorz-czajkowski:
Provide example and implementation in TAU.
See example in Appbar (Max-width)

grzegorz-czajkowski:
There is discrepancy in guide related to dropdown max-width.

In 13-4 Dropdown list (Max width & height) on page 270 , long content is presented as follows:
`Option11 Option11 Option11 Option...`

one line only and dots are added to notify user that content has been shorten. This is already implemented by https://github.com/Samsung/TAU/pull/828.

However, in Appbar More optionlist (Max width) on page 51,  long content is presented as follows:
```
Lorem ipsum dolor sit amet, consectetur
adipiscing elit, sed do eiusmod tempor...
```
which means, content should be presented in maxium two lines and have a dots for longer text.

@singa2000  and @lslachciak the question is:
1) should we support one line item or
2) should we support two lines item or
3) both

lslachciak:
@grzegorz-czajkowski I think both.
We have also two versions of Apbar - one line and two line

grzegorz-czajkowski:
This feature requires -webkit-line-clamp css property which is not supported in Tizen Emulator 5.5:
![emulator](https://user-images.githubusercontent.com/46523330/77900303-27f47780-727e-11ea-86b6-8924456dc0aa.png)
whereas on Desktop using Chromium | 80.0.3987.87 (Official Build) Built on Ubuntu i looks like expected:
![Chromium](https://user-images.githubusercontent.com/46523330/77900422-4e1a1780-727e-11ea-9266-76ee2663c924.png)

Using line-height and max-heigh = 2 * line-height causes bottom-padding overflow. In addition, we would add "..." if string is longer in js code.

I will try creating two spans within li element to just show example.

grzegorz-czajkowski:
Patch merged. Issue can be closed.

PR #856, opened 2020-03-30
Dropdown list (Mobile): two lines

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/836
[Problem] No support and example for option displayed in
	two lines.
[Solution] Add example and separate class.
[Remarks] This should be implemented using -webkit-line-clamp
	but this does not work on Tizen Emulator 5.5.
	TODO: add support for text ellipsis for second line manually.

### DIFF
--- a/examples/mobile/UIComponents/components/appbar/more-option-list-max-width.html
+++ b/examples/mobile/UIComponents/components/appbar/more-option-list-max-width.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta content="width=device-width, user-scalable=no" name="viewport" />
+	<link href="../../lib/tau/mobile/theme/default/tau.css" rel="stylesheet" />
+	<link href="../../css/style.css" rel="stylesheet" />
+	<script data-build-remove="false" src="../../lib/tau/mobile/js/tau.js"></script>
+</head>
+
+<body>
+	<div class="ui-page">
+		<header>
+			<div class="ui-appbar-left-icons-container">
+				<a href="#" class="ui-btn ui-btn-icon ui-btn-icon-back" data-style="flat" data-rel="back"></a>
+			</div>
+			<div class="ui-appbar-action-buttons-container">
+				<button class="ui-btn ui-btn-icon ui-btn-icon-more" data-style="flat" onClick="openDropdownMenu()"></button>
+				<select data-native-menu="false" id="dropdownmenu" style="display: none">
+					<option value="1">
+						Option1
+					</option>
+					<option value="2" class="ui-dropdown-two-lines">
+						Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam consectetur venenatis blandit. Praesent vehicula, libero non pretium vulputate, lacus arcu facilisis lectus, sed feugiat tellus nulla eu dolor. Nulla porta bibendum lectus quis euismod. Aliquam volutpat ultricies porttitor. Cras risus nisi, accumsan vel cursus ut, sollicitudin vitae dolor. Fusce scelerisque eleifend lectus in bibendum. Suspendisse lacinia egestas felis a volutpat.
+					</option>
+					<option value="3">
+						Option3
+					</option>
+					<option value="4">
+						Option4
+					</option>
+				</select>
+			</div>
+		</header>
+
+		<div class="ui-content">
+		</div>
+
+		<script>
+			function openDropdownMenu() {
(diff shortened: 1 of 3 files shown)


examples/mobile/UIComponents/components/appbar/more-option-list-max-width.html
```diff
@@ -0,0 +1,49 @@
...
+			<div class="ui-appbar-left-icons-container">
+				<select data-native-menu="false">
+					<option data-placeholder="true" value="first-line" data-hide-placeholder-menu-items="false">
+						Open
+					</option>
+					<option value="1">
+						Option1
+					</option>
+					<option value="2" class="ui-dropdown-two-lines">
+						Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam consectetur venenatis blandit. Praesent vehicula, libero non pretium vulputate, lacus arcu facilisis lectus, sed feugiat tellus nulla eu dolor. Nulla porta bibendum lectus quis euismod. Aliquam volutpat ultricies porttitor. Cras risus nisi, accumsan vel cursus ut, sollicitudin vitae dolor. Fusce scelerisque eleifend lectus in bibendum. Suspendisse lacinia egestas felis a volutpat.
```
DohyungLim, line 24: Can't it be changed to 2 lines?
The code width is too long.

examples/mobile/UIComponents/components/appbar/more-option-list-max-width.html
```diff
@@ -0,0 +1,49 @@
...
+			</div>
+			<div class="ui-appbar-action-buttons-container">
+				<button class="ui-btn ui-btn-icon ui-btn-icon-more" data-style="flat"></button>
+			</div>
+		</header>
+		<div class="ui-content">
+		</div>
+	</div>
+	<!-- /page -->
+	<script>
```
DohyungLim: Remove unnecessary lines.

src/css/profile/mobile/common/dropdownmenu.less
```diff
@@ -256,11 +256,18 @@
 			.font(regular);
 			display: block;
 			position: relative;
-			white-space: nowrap;
 			overflow: hidden;
 			text-overflow: ellipsis;
 			color: var(--dropdown-menu-options-color);
-
+			&.ui-dropdown-two-lines {
+				max-height: 2em;	// FIXME: max-height should be 2 * line-height but gives more than two lines.
```
DohyungLim, line 263: I think it is good to remove the FIXME line. or create an issue and delete it.

and is it correct to use em units?

src/css/profile/mobile/common/dropdownmenu.less
```diff
@@ -256,11 +256,18 @@
...
+				line-height: 1.4em;
+				// FIXME: add support for text ellipsis for second line manually since
+				//		  -webkit-line-clamp does not work on Tizen Emulator 5.5.
```
DohyungLim, line 266: ditto.

DohyungLim (commented):
4 comment.

lslachciak (commented):
I think more option list should be showed in reaction on the three dots press not Open button
See 1-4 More Option

grzegorz-czajkowski:
@lslachciak yes, you're right. To achieve, placeholder of dropdown should accept button styles but currently they are ignored and span element is created. Alternatively, from placeholder we should create a button instead of span.
I think we should manually register onClick event from "More option" and dynamically create Dropdown menu widget using tau.widget.DropdownMenu but I am facing errors. Need to check this.

examples/mobile/UIComponents/components/appbar/more-option-list-max-width.html
grzegorz-czajkowski: This is expected. User gives long text and widget formats it into two lines.

examples/mobile/UIComponents/components/appbar/more-option-list-max-width.html
grzegorz-czajkowski: Done.

src/css/profile/mobile/common/dropdownmenu.less
grzegorz-czajkowski: I am not sure if we should create issue for this since visual effect is fine. My intention of adding FIXME is to find better way to display text in two lines (for example, -line-clamp if it's supported in TIZEN).

Using em units is correct here.

grzegorz-czajkowski:
Added Dropdown Menu as part of header and set its "display: none" not to show it. After clcik on "More Options" we manually call show() method to display it.

lslachciak (approved):

Issue #852, opened 2020-03-30
Cotrollers --> Checkbox (List selected) -->Bottom bar show up


lmslachciak:
- showing the BottomBar if occures selected items

TomaszLukawskiSam:
Patch has been merged. The issue can be close.

lslachciak:
works fine.
Additionally I created new issue for wrong move icon in bottom bar
https://github.com/Samsung/TAU/issues/884

PR #864, opened 2020-03-31
AppBar: added hide/show feature for bottom bar on multiple selection list

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/852
[Problem] Bottom Bar should be hidden when no items selected on
 multiple selection list with "all" option
[Solution]
 - AppBar wigdet has beeen modified
 - new css style "ui-hidde" has been added for bottom bar

### DIFF
--- a/src/css/profile/mobile/common/bottombar.less
+++ b/src/css/profile/mobile/common/bottombar.less
@@ -25,6 +25,9 @@
         &.ui-bottom-bar-icons {
             padding: 0 10 * @px_base;
         }
+        &.ui-hidden {
+            display: none;
+        }
     }
 }
 
--- a/src/js/core/widget/core/Appbar.js
+++ b/src/js/core/widget/core/Appbar.js
@@ -37,7 +37,7 @@
 			//>>excludeEnd("tauBuildExclude");
 			var utilsObject = ns.util.object,
 				utilsEvents = ns.event,
-				selectors = ns.util.selectors,
+				utilSelectors = ns.util.selectors,
 				Page = ns.widget.core.Page,
 				min = Math.min,
 				max = Math.max,
@@ -50,7 +50,8 @@
 						leftIconsContainer: null,
 						actionButtonsContainer: null,
 						page: null,
-						selectAll: null
+						selectAll: null,
+						bottomBar: null
 					},
 					self._appbarState = states.COLLAPSED;
 					self._dragStartingHeight = 0;
@@ -70,7 +71,9 @@
 					dragging: classPrefix + "-dragging",
 					controlsContainer: classPrefix + "-controls-container",
 					expandedTitleContainer: classPrefix + "-expanded-title-container",
-					selectAll: "ui-label-select-all"
+					selectAll: "ui-label-select-all",
+					bottomBar: "ui-bottom-bar",
+					hidden: "ui-hidden"
 				},
 				states = {
 					EXPANDED: "EXPANDED",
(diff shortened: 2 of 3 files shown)


src/js/core/widget/core/Appbar.js
```diff
@@ -397,6 +404,34 @@
...
+			 * Toggle visibility of BottomBar
+			 * @method _toggleSelectAll
+			 * @param {boolean} visible
+			 * @member ns.widget.core.Appbar
+			 * @protected
+			 */
+			prototype._toggleBottomBar = function (visible) {
+				var bottomBar = this._ui.bottomBar;
+
+				if (visible) {
```
korneliakobiela: What do you think about using `classList.toggle(classes.hidden, true)` instead of this condition expression?

src/css/profile/mobile/common/bottombar.less
```diff
@@ -25,6 +25,9 @@
         &.ui-bottom-bar-icons {
             padding: 0 10 * @px_base;
         }
+        &.ui-hidden {
+            display: none;
+        }
```
HunseopJeong, line 30: How about moving the 'ui-hiddent' to a more common place? such as 'common.less'?
It seems to be more reusable.

TomaszLukawskiSam: I remember that general selector is often too weak.

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: Actually, I can use `toggle`. I usually wonder if the system supports this, but I checked that we use this in several places. I'll change this.

src/js/core/widget/core/Appbar.js
```diff
@@ -397,6 +404,28 @@
 				}
 			};
 
+			/**
+			 * Toggle visibility of BottomBar
+			 * @method _toggleSelectAll
```
HunseopJeong: Method name is wrong.

src/js/core/widget/core/Appbar.js
```diff
@@ -397,6 +404,28 @@
...
+
+			/**
+			 * Method check if page has listview with checkbox in state checked
+			 * @method _isCheckedCheckbox
+			 * @member ns.widget.core.Appbar
+			 * @return {boolean}
+			 * @protected
+			 */
+			prototype._isCheckedCheckbox = function () {
+				return !!this._ui.page.querySelector(".ui-listview input[type='checkbox']:checked");
```
DohyungLim: It is recommended to check @HunseopJeong comments.
I agree with his opinion.

`".ui-listview input[type='checkbox']:checked" can select all checked input element in the listview.`
`".ui-listview > input[type='checkbox']:checked" seems to be more proper in this situation.`

DohyungLim (commented):
1 comment.

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: I came to the conclusion that this method is not needed.

TomaszLukawskiSam: :). Mistake. This method is needed. Done

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: Method has been removed.

HunseopJeong (approved):
image
It works correctly to me.

PR #869, opened 2020-04-01
[mobile] fixup, dialog sample Uncaught TypeError fixed

DohyungLim:
- Uncaught TypeError: Cannot read property 'getElementById' of undefined
- Add document and tau to function parameters.


singa2000 (approved):
It's good

PR #882, opened 2020-04-02
Move rotary event to the wearable profile

HunseopJeong:
[Issue] N/A
[Problem] Rotary event is only for the wearable profile
[Solution] Move the code to the wearable profile


TomaszLukawskiSam (approved):
Verified. Works good. Ok.

singa2000 (approved):
It's good

PR #883, opened 2020-04-02
Spin (mobile): Applied UX guide for the Spin widget

HunseopJeong:
[Issue] N/A
[Problem] Default style for the spin was suitable for the wearable style
[Solution] Applied the UX guide for the Spin

### DIFF
--- a/src/js/core/widget/core/Spin.js
+++ b/src/js/core/widget/core/Spin.js
@@ -74,7 +74,7 @@
 				 *  // the spin will rotate by 1 -> 0 -> 9 -> 0
 				 * @property {number} [options.duration=ROLL_DURATION] time of rotate to indicated value
 				 * @property {string} [options.direction="up"] direction of spin rotation
-				 * @property {string} [options.rollHeight="container"] size of frame to rotate one item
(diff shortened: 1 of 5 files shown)


HunseopJeong:
image
There is no detaul UX guide for the Spin.
I modified only the parts that needed to be changed as much as possible and kept many parts.

singa2000 (approved):
It looks good.

src/js/core/widget/core/Spin.js
```diff
@@ -189,6 +189,7 @@
 					item.style.transform = "translateY(-1000px)"; // move item from active area
 				}
 				item.style.opacity = change.opacity;
+				item.style.height = itemHeight + "px";
```
DohyungLim, line 192: This code is not available in wearable. What is the effect on the item?

DohyungLim (commented):
1 comment.

src/js/core/widget/core/Spin.js
HunseopJeong: I checked in wearable profile.
It does not set the their height if useer change the itemHeight value.
So I added it.

DohyungLim (approved):
LGTM.

PR #892, opened 2020-04-03
Update Dialogs GUI 1.7 version

DohyungLim:
- Adjust content spacing.
- Sample modification.

### DIFF
--- a/src/css/profile/mobile/common/popup.less
+++ b/src/css/profile/mobile/common/popup.less
@@ -6,25 +6,22 @@
 	position: absolute;
 	left: 0;
 	margin-bottom: 16 * @px_base;
+	margin-top: 32 * @px_base;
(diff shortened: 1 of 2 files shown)


src/css/profile/mobile/common/popup.less
```diff
@@ -202,16 +199,16 @@
 
 	.ui-popup-notitle {
 		@media (orientation: portrait) and (max-width: 479px) {
-			margin-top: 24 * @px_base;
```
singa2000, line 205: In case of, we don't need to check the max-width?
Doesn't it have a different value only depending on the orientation mode?

DohyungLim: agree. I modified.

done.

singa2000 (approved):
It looks good

PR #901, opened 2020-04-06
Spin (mobile): Apply default options for mobile profile

HunseopJeong:
[Issue] N/A
[Problem] Need to apply different options for each profile
[Solution]
1. Add a new options variable to apply options for mobile profile
2. Delete vclick event handlers and enable the spin by default
3. Modify the test values to fit with changed options


singa2000 (approved):
LGTM

DohyungLim (approved):
LGTM.

Issue #808, opened 2020-03-23
View controls -> Dropdown list -> wrong position if opened from scrolled element


grzegorz-czajkowski:
If list is scrolled, click on item which is still visible cause wrong dropdown list position.
![top_align](https://user-images.githubusercontent.com/46523330/77318745-dc861a80-6d0d-11ea-89ae-2261873122b4.png)

Similar issue
![bottom](https://user-images.githubusercontent.com/46523330/77318755-e0b23800-6d0d-11ea-9812-0e150fb240b5.png)

PR #902, opened 2020-04-06
Fix wrong position of dropdown list

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/808
[Problem] when list item is hidden partially dropdown list is not placed properly
[Solution] Unfocus element before focusing it so focus function can be effective

### DIFF
--- a/cSpell.json
+++ b/cSpell.json
@@ -132,6 +132,7 @@
         "beforecreate",
         "taufocus",
         "taublur",
+        "unfocus",
         "longpress",
         "pinchstart",
         "pinchmove",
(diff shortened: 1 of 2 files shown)


grzegorz-czajkowski (commented):
This fixes position of dropdown menu but scroll on the list happens. As we discussed, please share alternative solution where position of dropdown list is updated based on top position.

lslachciak:
@grzegorz-czajkowski I think I have alternative solution
see:
https://github.com/Samsung/TAU/pull/903 (in progress)

lslachciak:
closing this PR in favour to alternative

Issue #896, opened 2020-04-03
Dialog -> Time Picker (12h)


HunseopJeong:
Dialog -> Time Picker (12h)

Need to implement

HunseopJeong:
PR is merged.

Issue #900, opened 2020-04-06
Dialog -> TimePicker -> ContainerHeight


HunseopJeong:
![image](https://user-images.githubusercontent.com/29908741/78542983-ed1cb180-7832-11ea-89d2-73801aceef5a.png)

I am implementing the TimePicker but the height of Spin has wrong value in the popup.
I think following code get the wrong value from the element.
`core / Spin.js`
```
prototype._refresh = function () {
    var self = this;
   
    self._containerRect = self.element.getBoundingClientRect();
    // Should have a correct size of height in self._containerRect but it set with 0.
}
```

HunseopJeong:
PR was merged.

PR #904, opened 2020-04-07
TimePicker (mobile): Implement the TimePicker (12h)

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/896
[Problem] Mobile profile doesn't support the TimePicker
[Solution]
1. Implement the TimePicker (12h)
2. Add the example for the TimePicker (12h)
3. Add tests for the TimePicker

### DIFF
--- a/src/js/profile/mobile/widget/TimePicker.js
+++ b/src/js/profile/mobile/widget/TimePicker.js
@@ -1,0 +1,293 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd
+ *
+ * Licensed under the Flora License, Version 1.1 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://floralicense.org/license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* global define */
+/**
+ * # TimePicker Widget
+ * Shows a control that can be used to set hours and minutes.
+ * It support 12/24 hours format. It contains two inputs which control the values
+ *
+ * ## Default selectors
+ *
+ * Default selector for timepicker is class *ui-time-picker*
+ *
+ * ### HTML Examples
+ *
+ * #### 12 hours format
+ * To add a timepicker widget to the application, use the following code:
+ *
+ *      @example
+ *      <div class="ui-time-picker" data-format="12">
+ *
+ * #### 24 hours format
+ * To add a timepicker widget to the application, use the following code:
+ *
+ *      @example
+ *      <div class="ui-time-picker" data-format="24">
+ *
+ * @class ns.widget.mobile.TimePicker
+ * @since 1.2
+ * @component-selector .ui-time-picker
+ * @extends ns.widget.mobile.TimePicker
+ * @author Hunseop Jeong <hs85.jeong@samsung.com>
+ */
+(function (window, document, ns) {
+	"use strict";
+	//>>excludeStart("tauBuildExclude", pragmas.tauBuildExclude);
+	define(
+		[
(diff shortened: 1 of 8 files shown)


HunseopJeong:
Current snapshot of this PR
image

Remain Issue.
1) Spin has wrong height value (https://github.com/Samsung/TAU/issues/900)
2) The event is delivered through the popup.

src/js/profile/mobile/widget/TimePicker.js
```diff
@@ -0,0 +1,295 @@
...
+					MINUTE_CONTAINER: WIDGET_CLASS + "-container-minute",
+					FORMAT_CONTAINER: WIDGET_CLASS + "-container-format"
+				},
+				WIDGET_SELECTOR = "." + WIDGET_CLASS,
+				TimePicker = function () {
+					this.options = {
+						format: "12"
+					};
+
+					this._ui = {};
```
DohyungLim: this._ui is not required.

src/js/profile/mobile/widget/TimePicker.js
```diff
@@ -0,0 +1,295 @@
...
+				* @protected
+				*/
+			prototype._init = function () {
+				var self = this,
+					options = self.options,
+					initDate = new Date();
+
+				options.format = (options.format !== undefined) ? options.format : "12";
+
+				self._setValue(initDate);
```
DohyungLim: question.
when init, does setValue set to the current time?
Is there any reason to differentiate between setValue and setDateValue?

DohyungLim (commented):
2 comment.

src/js/profile/mobile/widget/TimePicker.js
TomaszLukawskiSam: the variable `initDate` seems redundant... `self._setValue(new Date);` seems enough

src/js/profile/mobile/widget/TimePicker.js
HunseopJeong: Right. I added it to use in next PR.
I will remove it now.

src/js/profile/mobile/widget/TimePicker.js
HunseopJeong: Yes. It set the value with current time.
I will extend the setValue to set the time with Date Object, Object or value.
Currently, I impelmented it to set with Date Object only.

src/js/profile/mobile/widget/TimePicker.js
```diff
@@ -0,0 +1,294 @@
...
+
+				self.option("format", options.format);
```
TomaszLukawskiSam, line 99: I don't see method `prototype._setFormat` which can be handle this call. Probably only effect will be `refresh` the widget.

src/js/profile/mobile/widget/TimePicker.js
```diff
@@ -0,0 +1,294 @@
...
+
+				element.appendChild(hourPicker);
+				element.appendChild(minutePicker);
+				element.appendChild(formatPicker);
+
+				return element;
+			};
+
+			/**
+				* Build Spin Widget for the time
```
TomaszLukawskiSam: Indents in jsDoc seems wrong

src/js/profile/mobile/widget/TimePicker.js
HunseopJeong: Okay. I will remove it.

src/js/profile/mobile/widget/TimePicker.js
HunseopJeong: I used it to add the 'data-format' attribute because we need a selector to distinguish between 12 and 24.

TomaszLukawskiSam: After this will be change also data-attribute=format... so ok

src/js/profile/mobile/widget/TimePicker.js
HunseopJeong: Okay. I will fix it!

TomaszLukawskiSam (approved):
I think the widget is defined very well.

singa2000 (approved):
It looks good.

Issue #894, opened 2020-04-03
Text Fields --> Text Field v.17 update


lmslachciak:

lslachciak:
PR merged

PR #907, opened 2020-04-07
Update line height for text fields according to 1.7 version of guides

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/894
[Problem] Wrong line height
[Solution] Apply line height according to the guideline

### DIFF
--- a/src/css/profile/mobile/common/textinput.less
+++ b/src/css/profile/mobile/common/textinput.less
@@ -103,7 +103,7 @@ textarea.ui-text-input {
 	box-sizing: border-box;
 	display: block;
 	width: 100%;
-	line-height: 54 * @unit_base;
+	line-height: 26 * @px_base;
 	.font(regular);
 	-webkit-text-fill-color: transparent;
 	font-size: @font_size_textinput_text;


TomaszLukawskiSam (approved):
Looks good to me.

src/css/profile/mobile/common/textinput.less
```diff
@@ -103,7 +103,7 @@ textarea.ui-text-input {
 	box-sizing: border-box;
 	display: block;
 	width: 100%;
-	line-height: 54 * @unit_base;
+	line-height: 26 * @sp_base;
```
HunseopJeong: Do we use the @sp_base for the 'line-height' not @px_base?

lslachciak: yes, in some places of guide I see font size values marked with sp so I assume all guide values refer to sp

lslachciak: @HunseopJeong you are right. For Line height we should use px

lslachciak:
updated PR according to @HunseopJeong comment

HunseopJeong (approved):
LGTM

TomaszLukawskiSam (approved):

Issue #700, opened 2020-03-04
User input elements -> Checkbox


lmslachciak:
User input elements -> Checkbox

grzegorz-czajkowski:
@singa2000 are you working on this? If not I can take care of this.

singa2000:
oh.. No, I'm checking another issue.
Would you take care of this, please?
Thanks.

grzegorz-czajkowski:
@lslachciak  Checkbox and Checkbox Small were updated according to guide. Howvere one example missing. Should we close this bug?

lslachciak:
I think we can close

Issue #905, opened 2020-04-07
Presentation views -> List -> Main Text Only -> ui-arc-listview-5 items have wrong size


grzegorz-czajkowski:
Items arround selected elements have wrong position, size and opacity.
Current state:
![Screenshot from 2020-04-06 14-14-15](https://user-images.githubusercontent.com/46523330/78634461-525cbb00-78a4-11ea-851e-ece4e2f32af1.png)

lslachciak:
Debugger shows me 36px guide shows 38 px for main item font size (List 03)

grzegorz-czajkowski:
font-size is already updated in https://github.com/Samsung/TAU/pull/887. I can see correct font size = 38px.

PR #908, opened 2020-04-07
Checkbox (Wearable):  update according to guideline

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/700
[Problem] Example at wrong path, update style for checkbox.
[Solution] Add example to correct path and update it
	according to guide.

### DIFF
--- a/examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
+++ b/examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
@@ -1,10 +1,9 @@
-<!DOCTYPE html>
 <html>
 
 <head>
 	<meta content="width=device-width, initial-scale=1, user-scalable=no" name="viewport" />
 	<title>
-		Wearable UI
+		Checkbox
 	</title>
 	<link href="../../../lib/tau/wearable/theme/default/tau.css" rel="stylesheet" />
 	<link href="../../../lib/tau/wearable/theme/default/tau.circle.css" media="all and (-tizen-geometric-shape: circle)" rel="stylesheet" />
@@ -17,41 +16,51 @@
 </head>
 
 <body>
-	<div class="ui-page" data-go-to-top-button="true">
-		<header class="ui-header">
-			<h2 class="ui-title">
-				Data usage
+	<div class="ui-page">
+		<header>
+			<h2>
+				Title
 			</h2>
 		</header>
 		<div class="ui-content">
 			<ul class="ui-listview">
 				<li class="li-has-checkbox">
 					<label>
-						Mobile Data
-						<input type="checkbox" />
+						Check on
+						<input type="checkbox" checked />
 					</label>
 				</li>
 				<li class="li-has-checkbox">
 					<label>
-						Mobile Data
-						<input type="checkbox" />
-					</label>
-				</li>
-				<li class="li-has-checkbox">
-					<label>
-						Mobile Data
-						<input type="checkbox" />
-					</label>
-				</li>
-				<li class="li-has-checkbox">
-					<label>
-						Mobile Data
+						Check off
 						<input type="checkbox" />
 					</label>
 				</li>
 			</ul>
 		</div>
-	</div>
+		<script>
+			var checkboxes = document.querySelectorAll("input[type='checkbox']");
+			checkboxes.forEach(function(checkbox) {
+				checkbox.addEventListener("change", function() {
+					var text = "Check ",
+						label = this.labels[0],
+						marqueeContent = label.querySelector(".ui-marquee-content") || label.querySelector(".ui-marquee");
+
+					text += this.checked ? "on" : "off";
+					if (marqueeContent) {
+						marqueeContent.innerHTML = text;
+					} else {
+						label.firstChild.nodeValue = text;
+					}
+				});
+			});
+		</script>
+	</div> <!-- ui-page -->
+
+	<script src="../../../js/app.js">
+	</script>
+	<script src="../../../js/lowBatteryCheck.js">
+	</script>
 </body>
 
 </html>
--- a/examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/index.html
+++ b/examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/index.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta content="width=device-width, initial-scale=1, user-scalable=no" name="viewport" />
+	<title>
+		Checkbox
+	</title>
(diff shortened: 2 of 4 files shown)


grzegorz-czajkowski:
If checkbox is not in the middle of screen, its size is wrong 50x50 but according to guide (html) should be 53x53. This size comes from transform operation calculated by JS. Similar issue for ui-arc-listview-5 is already reported https://github.com/Samsung/TAU/issues/905.

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/index.html
```diff
@@ -0,0 +1,45 @@
...
+				</li>
+
+				<!-- TODO: Add more examples -->
+
+			</ul>
+		</div>
+	</div>
+	<script src="../../../js/app.js">
+	</script>
+	<script src="../../../js/lowBatteryCheck.js">
```
lslachciak, line 41: could you make sure if this is JS is really needed?

lslachciak (changes requested):
this example doesn't work properly (check on/off is not changed) if accessed using menu items both on simulator and in browser.
If I open directly in browser example it works

see http://localhost:8080/examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/index.html
vs.
http://localhost:8080/examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
```diff
@@ -18,40 +17,46 @@
...
 					</label>
 				</li>
 			</ul>
 		</div>
 	</div>
+	<script src="../../../js/app.js">
+	</script>
+	<script src="../../../js/lowBatteryCheck.js">
+	</script>
+	<script>
```
lslachciak: the problem with not working script when accessed from main menu(about which I wrote in previous comment) may be that you are not including this inside of 
`<div class="ui-page" data-go-to-top-button="true">`

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
```diff
@@ -1,13 +1,12 @@
...
 <html>
 
 <head>
 	<meta content="width=device-width, initial-scale=1, user-scalable=no" name="viewport" />
 	<title>
-		Wearable UI
+		Checkbox
 	</title>
 	<link href="../../../lib/tau/wearable/theme/default/tau.css" rel="stylesheet" />
-	<link href="../../../lib/tau/wearable/theme/default/tau.circle.css" media="all and (-tizen-geometric-shape: circle)" rel="stylesheet" />
```
HunseopJeong, line 10: Don't remove the  'media="all and (-tizen-geometric-shape: circle)"'

src/css/profile/wearable/changeable/theme-circle/checkboxradio.less
```diff
@@ -186,12 +188,12 @@ input[type="radio"] {
 				width: 100%;
 
 				div + input {
-					margin-left: 10 * @unit_base;
+					margin-left: 8 * @px_base;
```
HunseopJeong, line 191: ditto.

src/css/profile/wearable/changeable/theme-circle/checkboxradio.less
```diff
@@ -165,6 +165,8 @@ input[type="radio"] {
 /** CheckBox / Radio button on arc-listview **/
 .ui-arc-listview-carousel-item.ui-listview {
 	li {
+		padding-left: 30 * @px_base;
+		padding-right: 16 * @px_base;
```
HunseopJeong, line 169: In wearable profile, we use the @unit_base by default.

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/index.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta content="width=device-width, initial-scale=1, user-scalable=no" name="viewport" />
+	<title>
+		Button
+	</title>
+	<link href="../../../lib/tau/wearable/theme/default/tau.css" rel="stylesheet" />
+	<link href="../../../lib/tau/wearable/theme/default/tau.circle.css" rel="stylesheet" />
```
HunseopJeong: ditto.

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
grzegorz-czajkowski: They cause wrong layout on Browser and we already started to get rid of them.

src/css/profile/wearable/changeable/theme-circle/checkboxradio.less
grzegorz-czajkowski: Recently we started using @px_base in Wearable profile as well in consistency with Mobile.

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
HunseopJeong: TAU currently supports Square and Circle types. If you delete this, it only applies to the Circle, so I don't think it is correct.
@lslachciak, Is it right?

src/css/profile/wearable/changeable/theme-circle/checkboxradio.less
HunseopJeong: Okay, I see.

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/index.html
grzegorz-czajkowski: Most examples are using. I think it's better to add it to new examples as well with consistency.

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
grzegorz-czajkowski: Yes, the problem is that while navigation only page is taken into account, others elements are not.

examples/wearable/UIComponents/contents/user-inputs-elements/checkbox/checkbox.html
lslachciak: we need to discuss this issue

grzegorz-czajkowski: Since it's not decided yet I restored media query.

lslachciak (approved):
works fine now

Issue #897, opened 2020-04-03
Dialog -> Time Picker (24h)


HunseopJeong:
Dialog -> Time Picker (12h)

Need to implement

HunseopJeong:
PR was merged.

PR #910, opened 2020-04-08
TimePicker (mobile): Support the 24h format type

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/897
[Problem] Need to implement the TimePicker(24h)
[Solution] Add an example and styles for the TimePicker (24h)

### DIFF
--- a/src/css/profile/mobile/common/timepicker.less
+++ b/src/css/profile/mobile/common/timepicker.less
@@ -22,6 +22,28 @@
 		}
 	}
 
+	&[data-format="24"] {
+		.ui-time-picker-container {
+			width: 30%;
+
+			&-hour::after {
+				content: ":";
+				height: 164 * @px_base;
+				width: 40% / 3;
+				font-size: 32 * @sp_base;
+				color: var(--primary-dark-color);
+				position: absolute;
+				display: inline-flex;
+				align-items: center;
+				justify-content: center;
+			}
+
+			&-format {
+				display: none;
+			}
+		}
+	}
+
 	.ui-spin {
 		width: 100%;
 		height: 100%;
--- a/examples/mobile/UIComponents/components/dialogs/picker.html
+++ b/examples/mobile/UIComponents/components/dialogs/picker.html
@@ -34,7 +34,7 @@
 					</a>
 				</li>
 				<li class="ui-li-anchor">
-					<a data-inline="true">
+					<a data-position-to="window" data-rel="popup" href="time-picker-dialog-24h">
 						Time picker Dialog (24h)
 					</a>
 				</li>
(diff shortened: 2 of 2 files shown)


singa2000 (approved):
It's good

src/css/profile/mobile/common/timepicker.less
```diff
@@ -22,6 +22,28 @@
...
 	}
 
+	&[data-format="24"] {
+		.ui-time-picker-container {
+			width: 30%;
+
+			&-hour::after {
+				content: ":";
+				height: 164 * @px_base;
+				width: 40 / 3%;
```
TomaszLukawskiSam: It looks incorrect... is this works?

HunseopJeong: Yes,, it works,, Wouldn't it be strange to use 'calc'?
According to guide(4-15), Time Spin has 30% for the width. So, the remaining area(40%) is divided into three.

TomaszLukawskiSam: I don't see the `calc` and I don't understand this calculation.

HunseopJeong: Yes, it is just the 40 over 3. 40/3 = 13.333333 %.

TomaszLukawskiSam: maybe you wanted to write:
`calc(40% / 3)`

HunseopJeong: Right.
Done.

src/css/profile/mobile/common/timepicker.less
```diff
@@ -22,6 +22,28 @@
...
 	}
 
+	&[data-format="24"] {
+		.ui-time-picker-container {
+			width: 30%;
+
+			&-hour::after {
+				content: ":";
+				height: 164 * @px_base;
+				width: calc(40% / 3);
```
TomaszLukawskiSam: I'm not sure your intention. You wanted calc to be done in `less` or `css`?
For `less` will be enough `40% / 3` but in css `~"calc(40% / 3)"`

TomaszLukawskiSam: I think it was typo

HunseopJeong: Okay. I will recover it to done it in `less`.

src/css/profile/mobile/common/timepicker.less
```diff
@@ -22,6 +22,28 @@
...
 	}
 
+	&[data-format="24"] {
+		.ui-time-picker-container {
+			width: 30%;
+
+			&-hour::after {
+				content: ":";
+				height: 164 * @px_base;
+				width: 40% / 3;
```
DohyungLim, line 32: Is the width of the specification document divided by 3?

I'm worried if the value after the decimal point is fine even if the width is not correct.

DohyungLim (commented):
1 comment

src/css/profile/mobile/common/timepicker.less
HunseopJeong: I think there is no way to calculate the correct size for the width.

PR #916, opened 2020-04-08
[wearable] Navigation elements - correct menu structure and examples

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/915
[Problem] Menu structure not in accordance to guideline
[Solution] Correct mennu structure. Remove unneeded examples from menu.


examples/wearable/UIComponents/contents/navigationelements/index.html
```diff
@@ -31,18 +31,13 @@ <h2 class="ui-title">
...
+						Scroll bar
 					</a>
 				</li>
 				<li>
-					<a href="fastscroll.html">
-						Fast Scroll
-					</a>
-				</li>
-				<li>
-					<a href="sectionchanger/index.html">
```
grzegorz-czajkowski, line 44: Shouldn't we remove this file? Is it used somewhere else?

lslachciak: no it is not used, I don't see section changer in guides too. 
But we need to keep backward apps compatibility. So as for now I would leave files and e.g. move them to some section in future (e.g. etc?)

grzegorz-czajkowski: Yes, I think it's better to move it to etc since not used files are usually candidates to remove.

lslachciak: ok, done

grzegorz-czajkowski (approved):
Looks good to me.

PR #929, opened 2020-04-13
[mobile] Apply opacity to the disable day of the calendar

DohyungLim:
- prev month disable day : opacity 10%
- next month disalbe day : opacity 40%
- current month active day : opacity 100%


singa2000 (approved):
It's good

Issue #938, opened 2020-04-14
Move unneeded items to Etc


lmslachciak:
Move items not present in guides  to Etc

PR #939, opened 2020-04-14
[wearable] Move unneeded items to Etc

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/938
[Problem] Items not present in guide are available in main menu
[Solution] Move items to Etc. Update paths.


grzegorz-czajkowski (approved):
Woks fine. During verification I notice that Circular SectionChanger works in the same way as horizontal one. Could you check this on your side as well?

hs85jeong (approved):
I know it's moving files, but it would be nice to change the file name to kebab style while doing the moving. It is not an important issue.

lslachciak:
> 
> 
> I know it's moving files, but it would be nice to change the file name to kebab style while doing the moving. It is not an important issue.

We need to change this globally so I propose to do this in separate issue
https://github.com/Samsung/TAU/issues/940

lslachciak:
> Woks fine. During verification I notice that Circular SectionChanger works in the same way as horizontal one. Could you check this on your side as well?

@lslachciak I will verify examples separately

lslachciak:
@grzegorz-czajkowski circular and horizotal section changer looks the same, but they works in different way. From the last section you can go to the first one in case of circular. In case of horizontal you just have edge effect. Verified on latest HEAD (including this PR)

lslachciak:
image

lslachciak:
image

grzegorz-czajkowski:
@lslachciak I missed this. Thanks for check and explanation.

Issue #878, opened 2020-04-01
View Controls --> Main Tab


lmslachciak:

TomaszLukawskiSam:
@lslachciak 
The patch has already merged. I think we can close the issue.

lslachciak:
right, closing issue

PR #943, opened 2020-04-15
MainTab: implementation of widget

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/878
[Problem] Main Tab widget is not implemented
[Solution]
 - added example in UIComponents
 - added new MainTab widget based on TabBar
 - changes in tests
 - added new task 'test:karma-single-test' to package.json
 - README update

[Known issues]
 - https://github.com/Samsung/TAU/issues/985

### DIFF
--- a/examples/mobile/UIComponents/components/view-controls/main-tab/main-tab-3-items.html
+++ b/examples/mobile/UIComponents/components/view-controls/main-tab/main-tab-3-items.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta content="width=device-width, user-scalable=no" name="viewport" />
+	<link href="../../../lib/tau/mobile/theme/default/tau.css" rel="stylesheet" />
+	<link href="../../../css/style.css" rel="stylesheet" />
+	<script data-build-remove="false" src="../../../lib/tau/mobile/js/tau.js">
+	</script>
+</head>
+
+<body>
+	<div class="ui-page">
+		<link href="./main-tab-sample.css" rel="stylesheet" />
+		<div class="ui-content">
+			<!-- this page will be replaced by page from link target of active tab -->
+		</div>
+		<div class="ui-main-tab ui-active">
+			<ul>
+				<li class="ui-li-anchor">
+					<a href="./one.html">
+						One
+					</a>
+				</li>
+				<li class="ui-li-anchor">
+					<a class="ui-tab-active" href="./two.html">
+						Two
+					</a>
+				</li>
+				<li class="ui-li-anchor">
+					<a href="./three.html">
+						Three
+					</a>
+				</li>
+			</ul>
+		</div>
+	</div>
+</body>
+
+</html>
--- a/examples/mobile/UIComponents/components/view-controls/main-tab/one-appbar.html
+++ b/examples/mobile/UIComponents/components/view-controls/main-tab/one-appbar.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta content="width=device-width, user-scalable=no" name="viewport" />
+	<link href="../../../lib/tau/mobile/theme/default/tau.css" rel="stylesheet" />
+	<link href="../../../css/style.css" rel="stylesheet" />
+	<script data-build-remove="false" src="../../../lib/tau/mobile/js/tau.js">
+	</script>
+</head>
+
+<body>
+	<div class="ui-page">
+		<link href="./main-tab-sample.css" rel="stylesheet" />
+		<header>
+			<div class="ui-appbar-title-container">
+				<span class="ui-appbar-title">One</span>
+			</div>
+		</header>
+		<div class="ui-content">
+			<div class="ui-content-area app-full-height">
+			</div>
+		</div>
+	</div>
+</body>
+
+</html>
--- a/src/js/core/widget/core/MainTab.js
+++ b/src/js/core/widget/core/MainTab.js
@@ -1,0 +1,264 @@
+/*global ns, define */
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd
+ *
+ * Licensed under the Flora License, Version 1.1 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://floralicense.org/license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*jslint nomen: true, plusplus: true */
+/**
+ * #Main Tab
+ * Main Tab component is dedicated to control page change in app.
+ * The component is positioned always on bottom of the screen.
+ *
+ * @class ns.widget.core.MainTab
+ * @extends ns.widget.BaseWidget
+ * @author Tomasz Lukawski <t.lukawski@samsung.com>
+ */
+(function (document, ns) {
+	"use strict";
+	//>>excludeStart("tauBuildExclude", pragmas.tauBuildExclude);
+	define(
+		[
+			"../../engine",
+			"../../event",
+			"../../util/selectors",
+			"../BaseWidget",
+			"../core",
+			"./Marquee"
+		],
+		function () {
+			//>>excludeEnd("tauBuildExclude");
+			var BaseWidget = ns.widget.BaseWidget,
+				engine = ns.engine,
+				utilSelectors = ns.util.selectors,
+				FILE_REGEXP = /[^\/]+$/,
+				classes = {
+					/**
+					 * Standard widget
+					 * @style ui-main-tab
+					 * @member ns.widget.core.MainTab
+					 */
+					MAIN_TAB: "ui-main-tab",
+					/**
+					 * Set Main Tab widget as visible
+					 * @style ui-main-tab-visible
+					 * @member ns.widget.core.MainTab
+					 */
+					VISIBLE: "ui-main-tab-visible",
+					/**
+					 * Set Tab as active
+					 * @style ui-tab-active
+					 * @member ns.widget.core.MainTab
+					 */
+					ACTIVE_TAB: "ui-tab-active"
+				},
+				MainTab = function () {
+					this._ui = {};
+				},
+				prototype = new BaseWidget();
+
+			MainTab.classes = classes;
+			MainTab.prototype = prototype;
+
+			/**
+			 * Build MainTab component
+			 * @method _build
+			 * @param {HTMLElement} element
+			 * @return {HTMLElement} Returns built element
+			 * @member ns.widget.core.MainTab
+			 * @protected
+			 */
+			prototype._build = function (element) {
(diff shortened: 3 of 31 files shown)


lslachciak (commented):
tab one always has different look than others tab (header is visible)
is this intended?
vs

TomaszLukawskiSam:
> tab one always has different look than others tab (header is visible)
> is this intended?
> image
> vs
> image

Yes. Now, this is intended. There is AppBar for check widget behavior with different page content.

examples/mobile/UIComponents/components/view-controls/main-tab/main-tab-3-items.html
```diff
@@ -0,0 +1,42 @@
...
+	</script>
+</head>
+
+<body>
+	<div class="ui-page">
+		<link href="./main-tab-sample.css" rel="stylesheet" />
+		<div class="ui-content">
+			<div class="ui-content-area app-full-height">
+				One
+			</div>
```
HunseopJeong: Is it also fill with content automatically?
`<!-- this content will be fill by content from link target of active tab -->"`

Did you set it up on purpose?

src/css/profile/mobile/common/oneui-common.less
```diff
@@ -65,8 +66,5 @@
     --expander-color: @expander-color;
     --spin-item-opacity: @spin-item-opacity;
     --grid-selection-color: fade(@_black, 30%);
-    --calendar-weekend-day-color: @calendar-weekend-day-color;
-    --calendar-weekend-color: @calendar-weekend-color;
-    --calendar-text-color: @calendar-text-color;
-    --calendar-arrow-color: @calendar-arrow-color;
-    --calendar-select-text-color: @calendar-select-text-color;
```
HunseopJeong: Is it necessary to remove this part?

src/js/core/widget/core/Page.js
```diff
@@ -720,6 +728,22 @@
 				self._setContent(element, self.options.content);
 			};
 
+			/**
+			 * Method try to find mainTab.
+			 * @method _findMainTab
+			 * @param {HTMLElement} element
```
HunseopJeong: Seems a no parameter

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,239 @@
...
+					contentArea = page.querySelector(".ui-content-area");
+					contentArea.classList.add("ui-content-area-disabled-top-rounding");
+				}
+			};
+
+			/**
+			 * Destroy MainTab component
+			 * @method _destroy
+			 * @param {HTMLElement} element
+			 * @return {HTMLElement} Returns built element
```
HunseopJeong: No params, return.

src/css/profile/mobile/common/oneui-common.less
TomaszLukawskiSam: I don't understand why this change is here. I think this is wrong effect of merge conflicts resolving by me. I'll fix it.

examples/mobile/UIComponents/components/view-controls/main-tab/main-tab-3-items.html
TomaszLukawskiSam: Done. Thank you.

src/css/profile/mobile/common/oneui-common.less
TomaszLukawskiSam: Fixed. Thank you.

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/Page.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,237 @@
...
+						} else {
+							// active page is out of Main Tab (hide MainTab)
+							element.classList.remove(classes.ACTIVE);
+						}
+					}
+				}
+			};
+
+			/**
+			 * Disable top round in content area
```
HunseopJeong, line 224: Add a method name

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,237 @@
...
+				var self = this,
+					target = event.target,
+					element = self.element,
+					links = self._ui.links,
+					router,
+					activeLink,
+					url,
+					uri;
+
+				if (self._ui.parentPage === event.target) {
```
HunseopJeong: Use target instaed of the event.target

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,237 @@
...
+					// try to find active tab
+					activeLink = links.filter(function (link) {
+						return link.classList.contains(classes.ACTIVE_TAB);
+					})[0] || element.querySelector("a");
```
HunseopJeong, line 187: Isn't there a possibility that activeLink could be null?

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,237 @@
...
+							return link.getAttribute("href").indexOf(uri) > -1;
+						})[0];
+						if (activeLink) { // current active page is still in scope of MainTab
+							if (!activeLink.classList.contains(classes.ACTIVE_TAB)) {
+								links.forEach(function (link) {
+									link.classList.toggle(classes.ACTIVE_TAB, (link.getAttribute("href").indexOf(uri) > -1));
+								});
+							}
+							// Content area without appbar has disabled top rounds on MainTab
+							self._disableTopRounds(event.target);
```
HunseopJeong: ditto.

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,237 @@
...
+
+				// change parent to pageContainer instead of page
+				pageContainer = utilSelectors.getClosestBySelector(element, ".ui-page-container");
+				pageContainer.appendChild(element);
+
+				return element;
+			};
+
+			/**
+			 * Handle events
```
HunseopJeong, line 129: Need a @method

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: This is possible if developer didn't add any tab to MainTab, maybe it is actually better to protect it.

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: Protection has been added.

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: Done

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: Done.

examples/mobile/UIComponents/components/view-controls/main-tab/one-appbar.html
```diff
@@ -0,0 +1,27 @@
...
+<body>
+	<div class="ui-page">
+		<link href="./main-tab-sample.css" rel="stylesheet" />
+		<header>
+			<div class="ui-appbar-title-container">
+				<span class="ui-appbar-title">One</span>
+			</div>
+		</header>
+		<div class="ui-content">
+			<div class="ui-content-area app-full-height">
```
HunseopJeong, line 21: How about adding the app-full-height to the "core.less"?
I think it is a necessary style.
I think 'fill-height' is better than 'app-full-height'.

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,264 @@
...
+			 * @param {Event} event
+			 * @method handleEvent
+			 * @protected
```
HunseopJeong, line 132: handleEvent is `@public` method.

src/js/core/widget/core/MainTab.js
```diff
@@ -0,0 +1,264 @@
...
+					 * @style ui-main-tab
+					 * @member ns.widget.core.MainTab
+					 */
+					MAIN_TAB: "ui-main-tab",
+					/**
+					 * Set Main Tab widget as active (visible)
+					 * @style ui-main-tab-active
+					 * @member ns.widget.core.MainTab
+					 */
+					ACTIVE: "ui-active",
```
lslachciak: ui-main-tab-active or ui-active?
in documentation you use ui-main-tab-active which I personally prefer

lslachciak (commented):
small comment regarding API in code, also in night theme I can see additional white line on bottom not existing in guide
image

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: This is intended. This method is protected in widget context even that its name doesn't fit to name convention. The 'handleEvent' is special method for object which listens on events (https://developer.mozilla.org/en-US/docs/Web/API/EventListener/handleEvent). I cannot prefixed this method like "_handleEvent" and I cannot treated this method as public.

Second thing, we don't mark public method by `@public`.  We use only `@protected` or `@private` in jsDoc

TomaszLukawskiSam:
> small comment regarding API in code, also in night theme I can see additional white line on bottom not existing in guide

In my opinion it is out of scope my change but I remember this border in guideline for content area... This is possible that it was changed recently, I'm not sure.

src/js/core/widget/core/MainTab.js
TomaszLukawskiSam: The "ui-active" better looks in samples for me :) ... but you are right we should take account whole name convention. If we have "ui-page-active" we should use "ui-main-tab-active". This will be part of API so this is important.

TomaszLukawskiSam: After consideration, this class is use only internal. This is not part of API only this widget can use it. This class makes MainTab visible so I changed this name to ".ui-main-tab-visible".

examples/mobile/UIComponents/components/view-controls/main-tab/one-appbar.html
TomaszLukawskiSam: I'm not sure if the adding this to framework is good idea. We can do this in separate change.

About name convention... I want to distinct framework classes and classes specific for app.

src/js/core/widget/core/MainTab.js
HunseopJeong: Okay. If you are right, we need to change the all @public to the @protected or @private.
I marked the handleEvent as @public before because the handleEvent was marked as @public.

examples/mobile/UIComponents/components/view-controls/main-tab/one-appbar.html
HunseopJeong: However, it is inefficient to add additional css in all examples for this small one. 
I think you can add the style you need. It is my personal option. Anyway, I got it.

HunseopJeong (approved):
LGTM

lslachciak (approved):

Issue #730, opened 2020-03-05
Presentation View -> List ->  main text only


lmslachciak:
Presentation View -> List ->  main text only

lslachciak:
5 list item example
items in guideline are bigger

grzegorz-czajkowski:
Patch merged. Issue can be closed.

PR #945, opened 2020-04-15
Remove unneeded 5 items example

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/730

[Problem] Example for 5 items list already exist
in Presentation Views -> List -> Main text only

[Solution] Remove unneeded example


grzegorz-czajkowski (approved):
Correct. This is duplication of List -> Main text only.

PR #946, opened 2020-04-15
Move Load more contents example to proper place

lslachciak:
[Issue] N/A
[Problem] Load more contents example is in dfferent place
of guide structure
[Solution] Move example to Etc, where it should be


grzegorz-czajkowski (approved):
Works fine.

PR #961, opened 2020-04-17
Fix a problem that opacity is not applied on emulator

HunseopJeong:
[Issue] N/A
[Problem] Opacity effect is not applied when using the percent
[Solution] Change the opacity value to use the floating value

### DIFF
--- a/src/css/profile/mobile/common/spin.less
+++ b/src/css/profile/mobile/common/spin.less
@@ -23,7 +23,7 @@
 		.font(medium);
 
 		&-selected {
-			opacity: 100% !important;
+			opacity: 1 !important;
 		}
 	}
 	&-enabling {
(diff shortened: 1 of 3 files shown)


TomaszLukawskiSam (approved):
Looks good to me.

singa2000 (approved):
It's good working

TomaszLukawskiSam:
@HunseopJeong 
Hi,
This problem intrigued me and I found convertion method in `less` http://lesscss.org/functions/#misc-functions-unit
eg. `--opacity: unit(10%) / 100`
This allows to remains the same format of write a values for opacity

HunseopJeong:
@TomaszLukawskiSam,
Oh I see,, I will create the issue for this problem.
Thank you.

PR #963, opened 2020-04-17
TimePicker (mobile): Apply a two digits for minute picker

HunseopJeong:
[Issue] N/A
[Problem] Minute picker appears as a single digit
[Solution] Change option of the Minute Picker to use the two digits

### DIFF
--- a/src/js/profile/mobile/widget/TimePicker.js
+++ b/src/js/profile/mobile/widget/TimePicker.js
@@ -168,6 +168,7 @@
 				} else if (name === "minute") {
 					options.min = 0;
 					options.max = 59;
+					options.digits = 2;
 					spinContainer.classList.add(TimePicker.classes.MINUTE_CONTAINER);
 				}
 


src/js/profile/mobile/widget/TimePicker.js
```diff
@@ -252,7 +253,7 @@
 				ui.minuteInput.value = minutes;
 
 				spins.hour.value(hours);
-				spins.minute.value(minutes);
+				spins.minute.value((minutes < 10 ? "0" : "") + minutes);
```
TomaszLukawskiSam: Do you really needs this?

HunseopJeong: I refer to the TimePicker in wearable profile.
I will check it without it.

TomaszLukawskiSam: I have to verify it

HunseopJeong: @TomaszLukawskiSam,
You are right. It seems unnecessary.

TomaszLukawskiSam (approved):
Looks good to me.

singa2000 (approved):
It's good

PR #967, opened 2020-04-17
[wearable] Correct graph library path

lslachciak:
[Issue] N/A
[Problem] Bar graph not showing
[Reason] Incorrect graph library path
[Solution] Correct path

### DIFF
--- a/examples/wearable/UIComponents/contents/etc/graphs/bar_graph.html
+++ b/examples/wearable/UIComponents/contents/etc/graphs/bar_graph.html
@@ -28,9 +28,9 @@
 			 data-yinit="0"
 			 data-axis-x-type="order"
 			 data-axis-x-unit="s"></div>
-        </div><!-- /content -->
-		<script src="../../lib/tau/libs/tau.graph.js" type="text/javascript"></script>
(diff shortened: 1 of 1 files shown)


examples/wearable/UIComponents/contents/etc/graphs/bar_graph.html
```diff
@@ -29,7 +29,7 @@ <h1>Bar Graph</h1>
 			 data-axis-x-type="order"
 			 data-axis-x-unit="s"></div>
         </div><!-- /content -->
-		<script src="../../lib/tau/libs/tau.graph.js" type="text/javascript"></script>
+		<script src="../../../lib/tau/libs/tau.graph.js" type="text/javascript"></script>
```
grzegorz-czajkowski, line 32: Please check indents. Lokks like it should be aligned to script below since they are on the same parent (page).

lslachciak: fixed

grzegorz-czajkowski (approved):
Looks good to me.

PR #968, opened 2020-04-17
[wearable] Remove List from Etc menu

lslachciak:
[Issue] N/A
[Problem] List is already available in Presentation Views.
Not all list items are verified and in accordance to guideline.
[Solution] Remove List menu


grzegorz-czajkowski (approved):
I am fine with removing it just from menu. Every example from erc/list needs adjustments and fix before it will be placed into presentation-view/list.

PR #977, opened 2020-04-20
(tests) Fix for test of util/polar

TomaszLukawskiSam:
[Issue] N/A
[Problem] Util/polar test fails on latest browser
[Solution] Web browser behaviour has been changed for
 transformOrigin style and test fails

### DIFF
--- a/tests/js/core/util/polar/polar.js
+++ b/tests/js/core/util/polar/polar.js
@@ -146,10 +146,8 @@ function polarTests(polar) {
 
 		if (!window.navigator.userAgent.match("PhantomJS")) {
 			ok(path.classList.contains("ui-animated"), "animated class on path is correct set");
-			equal(path.style.transformOrigin, "120px 140px 0px", "style transform origin on path is correct set");
-		} else {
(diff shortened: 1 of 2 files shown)


HunseopJeong (approved):
LGTM

DohyungLim (approved):
looks good.

lslachciak (approved):
change seems ok, but I have one doubt, this test is executed only for mobile in case of phantom based test execution
https://travis-ci.org/github/Samsung/TAU/builds/677268976
do you know why it is not considered for wearable?

singa2000:
In case of tct, we have just mobile TCT.
Wearable don't have the TCT.

I have a question,
It it also apply in the tau_1.1 branch?

lslachciak:
@singa2000 I don't have concern regarding TCT but test executed on CI.
I think this change is also applicable to other branches too.

TomaszLukawskiSam:
> I have a question,
> It it also apply in the tau_1.1 branch?

I think this patch can be partially applied to tau_1.1 / tau_1.0 branch. We can copy there the fix for test.
but
we don't need change refers to karma test. Karma tests are implemented only on 1.2 branch.

Issue #971, opened 2020-04-18
TAU 1.1.8 (Tizen Studio 3.7) doesn't support number input fields


nemethmik:
Hello,
In the video I created the TAU UI Components sample project with Tizen Studio, and when I wanted to change the **input type** to **number** the application was broken.
[TAU 1.1.8 (Tizen Studio 3.7) doesn't support number input fields](https://youtu.be/a8TlFG23rkY)
In our application we have a number of simple screens with number input fields.
Thank you so much,
Miklos

HunseopJeong:
Hello, @nemethmik,
It was easy to understand that checked the issue with a video.
I will check the issue.

HunseopJeong:
@lslachciak,
Could you merge this PR(https://github.com/Samsung/TAU/pull/978) to all branches in next release?

lslachciak:
> 
> 
> @lslachciak,
> Could you merge this PR(#978) to all branches in next release?

Yes, I'm on it. 1.1 is done. 1.2 is in progress

lslachciak:
PR was merged to all branches

lslachciak:
@nemethmik appropriate changes were introduced into TAU 1.0, 1.1 and 1.2 and released 
https://github.com/Samsung/TAU/releases/tag/v.1.2.1
although process of incorporating changes into Tizen Studio can take some time (tests, etc..) and is out of scope of TAU issue

If you want I can prepare special release of TAU 1.1 too which has this change

PR #978, opened 2020-04-21
TextInput (wearable): Input element is broken with number type

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/971
[Problem] The number type of the input element is not working correctly
[Solution] Add a number type to the styles and add a condition to check
whether input element supports the selectionStart property

### DIFF
--- a/examples/wearable/UIComponents/contents/controls/text-input.html
+++ b/examples/wearable/UIComponents/contents/controls/text-input.html
@@ -20,8 +20,8 @@
 	<div class="ui-page" id="page-text-input">
 		<div class="ui-content text-input-content">
 			<input id="name" name="name" placeholder="Name" type="text" />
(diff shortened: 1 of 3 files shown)


HunseopJeong:
I think this PR needs to be applied from 1.0 to 1.2.

lslachciak (commented):
Could you also update example
https://github.com/Samsung/TAU/blob/tau_1.0/examples/wearable/UIComponents/contents/controls/text-input.html
so we can show that we support also number type
e.g. instead of phone number we can set year of birth

And I agree this change should be applied to all branches

HunseopJeong:
@lslachciak,
Yes, you are right. I also updated the example.
Done.

lslachciak (approved):
works well. can be merged

singa2000 (approved):
It looks good.

PR #979, opened 2020-04-21
List (mobile): Change some styles according to GUI guide

HunseopJeong:
[Issue] N/A
[Problem] Some styles are different from GUI guide
[Solution] Change some styles to align with GUI guide

### DIFF
--- a/src/css/profile/mobile/common/button.less
+++ b/src/css/profile/mobile/common/button.less
@@ -287,12 +287,6 @@ a.ui-btn {
 			width: 208 * @unit_base;
 		}
 
-		&::after {
-			width: 60 * @unit_base;
-			height: 60 * @unit_base;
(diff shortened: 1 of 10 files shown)


src/css/profile/mobile/common/button.less
```diff
@@ -287,12 +287,6 @@ a.ui-btn {
 			width: 208 * @unit_base;
 		}
 
-		&::after {
-			width: 60 * @unit_base;
-			height: 60 * @unit_base;
-			background-color: var(--button-icon-color);
-		}
-
```
HunseopJeong, line 295: is it neccessary part for the button?
It make some weird result in fab button.
image

HunseopJeong:
I'll divide it up for easy review.

PR #984, opened 2020-04-21
(SVACE) fix major issues (part 1)

grzegorz-czajkowski:
Contains the following commits:
 * BaseKeyboardSupport.js (SVACE): add return to filterNeighbors
 * BaseKeyboardSupport.js (SVACE): add conversion to Boolean
 * BaseKeyboardSupport (SVACE): remove unnecessary argument
 * Graph.js (SVACE): add return value for _build func

### DIFF
--- a/src/js/core/widget/BaseKeyboardSupport.js
+++ b/src/js/core/widget/BaseKeyboardSupport.js
@@ -662,11 +662,11 @@
 						// corner cases, when element positions overlap
 						if (focusableElement.distance === 0) {
 							if (direction === EVENT_POSITION.down) {
-								return currentLink.compareDocumentPosition(typeof focusableElement === element ? focusableElement : focusableElement.element) &
-										Node.DOCUMENT_POSITION_CONTAINED_BY;
+								return !!(currentLink.compareDocumentPosition(typeof focusableElement === element ? focusableElement : focusableElement.element) &
+										Node.DOCUMENT_POSITION_CONTAINED_BY);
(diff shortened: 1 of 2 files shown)


src/js/core/widget/BaseKeyboardSupport.js
```diff
@@ -662,11 +662,11 @@
 						// corner cases, when element positions overlap
 						if (focusableElement.distance === 0) {
 							if (direction === EVENT_POSITION.down) {
-								return currentLink.compareDocumentPosition(typeof focusableElement === element ? focusableElement : focusableElement.element) &
-										Node.DOCUMENT_POSITION_CONTAINED_BY;
+								return !!(currentLink.compareDocumentPosition(typeof focusableElement === element ? focusableElement : focusableElement.element) &
```
lslachciak, line 665: Do we always want o have boolean type?
According o func description it returns object
* @return {Object}

lslachciak (commented):
one doubt

src/js/core/widget/BaseKeyboardSupport.js
grzegorz-czajkowski: This documentation refers to outer function *getNeighborhoodLinks* whereas we are changing callback function for Array.filter where true/false are expected..

lslachciak: right, sorry I overlooked this

lslachciak (approved):

PR #986, opened 2020-04-21
Page (mobile): Fix to calculate the correct size of header

HunseopJeong:
[Issue] N/A
[Problem] The height of the content is not set properly
[Solution] Calculate to reflect the size of the margin


TomaszLukawskiSam (approved):
I know this issue. I agree with you.

Issue #980, opened 2020-04-21
GUI Feedback List


HunseopJeong:
- _List Content Area_
- _Bottom Spacing_
- _Bottom Spacing with FAB_
- _1 Line List_
- _2 Line List_
- _3 Line List_
- _Icon +Text List_
- Sub header / Text
- Sub header / Without chunking
- Expandable List
- Multiple Selection List / Basic list - Extended app bar
- Multiple Selection List / Basic list - App bar

Italic: complete

HunseopJeong:
Resovled

PR #987, opened 2020-04-21
List (mobile): Fix to apply GUI feedback

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/980
[Problem] Need to align with GUI guide
[Solution]
- List Content Area: box-shadow only applied to the light theme
- Bottom Spacing: Applied a masking to the ui-content
                  Wrong spacing of bottom can be fixed with #986
                  Remove a wrong style for the fab button

### DIFF
--- a/src/css/profile/mobile/common/core.less
+++ b/src/css/profile/mobile/common/core.less
@@ -134,6 +134,11 @@
 	overflow-y: visible;
 	overflow-x: hidden;
 	flex-shrink: 1;
+	mask-border: url(images/0_Round_corner/round.svg) 77;
+	mask-border-width: 26 * @px_base;
+	border-radius: 26 * @px_base;
+	box-sizing: border-box;
+	box-shadow: 0 0 0 0.25 * @px_base var(--content-area-line-color) inset;
 
 	&.ui-content-padding {
 		&, &.ui-scrollview-clip {
@@ -162,6 +167,12 @@
 		padding-bottom: 5 * @px_base;
 		margin-top: -4.25 * @px_base;
 		margin-left: 24 * @px_base;
+	}
+}
+
+body.ui-theme-dark {
+	.ui-content-area {
+		box-shadow: unset;
 	}
 }
 
--- a/src/css/profile/mobile/common/button.less
+++ b/src/css/profile/mobile/common/button.less
@@ -287,12 +287,6 @@ a.ui-btn {
 			width: 208 * @unit_base;
 		}
 
-		&::after {
-			width: 60 * @unit_base;
-			height: 60 * @unit_base;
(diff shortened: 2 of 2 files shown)


src/css/profile/mobile/common/core.less
```diff
@@ -134,6 +134,11 @@
 	overflow-y: visible;
 	overflow-x: hidden;
 	flex-shrink: 1;
+	mask-border: url(images/0_Round_corner/round.svg) 77;
```
singa2000, line 137: What's mean the "77" ?

src/css/profile/mobile/common/core.less
```diff
@@ -165,6 +170,12 @@
 	}
 }
 
+body.ui-theme-dark {
```
singa2000, line 173: Is there no need to consider light theme?
The followings code is for light theme?

	.ui-content-area {
		mask-border: url(images/0_Round_corner/round.svg) 77;
		mask-border-width: 26 * @px_base;
		border-radius: 26 * @px_base;
		box-sizing: border-box;
		box-shadow: 0 0 0 0.25 * @px_base var(--content-area-line-color) inset;
	}

src/css/profile/mobile/common/core.less
singa2000: In the GUI Guide, there are "top_left.svg" ,"top_right.svg","buttom_left.svg" "bottom_right.svg".
Do you use the  svg file?

HunseopJeong: Need to check that I just copy the style related with the mask from the 'ui-content-area'

src/css/profile/mobile/common/core.less
HunseopJeong: In 51 page, mask does not need to be applied to dark theme.

src/css/profile/mobile/common/core.less
HunseopJeong: 77 is for the "mask-border-slice"
```
<'mask-border-slice'>
The dimensions for slicing the source image into regions. Up to four values may be specified. See mask-border-slice.
```
round.svg is draw the circle(radius 78). So I can change it to the 78.

HunseopJeong: image
Oops,,, It is result when changing the slice value to the 78.
77 pixel seems a correct value

singa2000: ok. It's no problem.
Thanks for your checking.

src/css/profile/mobile/common/core.less
singa2000: ok. I see~

singa2000 (approved):
It looks good.

PR #988, opened 2020-04-22
Bottom bar : Fix to apply GUI feedback

singa2000:
[Issue] N/A
[Problem] Bottom bar is differenet with  GUI
[Solution] Bottom bar font is update

### DIFF
--- a/src/css/profile/mobile/common/bottombar.less
+++ b/src/css/profile/mobile/common/bottombar.less
@@ -10,13 +10,17 @@
         padding: 0 24 * @px_base;
     
         .ui-btn {
+            .font(medium);
             font-size: 18 * @sp_base;
+            text-align: center;
+            line-height: 36 * @unit_base;
             max-width: 100%;
             height: 56 * @px_base;
             &.ui-btn ~ .ui-btn {
                 margin-left: 0;
             }
             &.ui-btn-icon {
+                .font(regular);
                 font-size: 12 * @sp_base;
                 line-height: normal;
             }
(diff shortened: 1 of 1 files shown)


HunseopJeong (approved):
LGTM

src/css/profile/mobile/common/bottombar.less
```diff
@@ -33,7 +37,7 @@
     }
 }
 
-@media all and (min-width: 685px) and (min-height: 411px) {
+@media all and (min-width: 673px) and (min-height: 411px) {
     .ui-footer .ui-bottom-bar {
         margin: 0 auto;
         width: 75%;
```
TomaszLukawskiSam, line 43: I think you have to change also `min-width` to 625px, because there are also left/right paddings 2*24px

singa2000: Thanks~
I'll change it.

singa2000: &.ui-bottom-bar-icons {
            min-width: 665 * @px_base;
        }
I need also change 665->645 ?
In case of icon+text , left/right paddings 2*10?

TomaszLukawskiSam: yes.

TomaszLukawskiSam: but 673px - 2 * 10px = 653px

TomaszLukawskiSam (approved):
Looks good to me.

PR #990, opened 2020-04-22
On/Off switch : Fix to apply GUI feedback

singa2000:
[Issue] N/A
[Problem] On/Off switch margin is some different with GUI
[Solution] On/Off switch margin is update

### DIFF
--- a/src/css/profile/mobile/common/onoffswitch.less
+++ b/src/css/profile/mobile/common/onoffswitch.less
@@ -13,7 +13,7 @@
 
     &-input {
         position: absolute;
-        width: 35 * @px_base;
+        width: 37 * @px_base;
         height: 20.5 * @px_base;
         border-radius: 9.25 * @px_base;
         -webkit-appearance: none;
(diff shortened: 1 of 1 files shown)


src/css/profile/mobile/common/onoffswitch.less
```diff
@@ -13,7 +13,7 @@
 
     &-input {
         position: absolute;
-        width: 35 * @px_base;
+        width: 37 * @px_base;
```
HunseopJeong, line 16: As I know as, 35 is correct size for the checkbox because it has 1px border.
@TomaszLukawskiSam, @lslachciak,
Isn't it?

TomaszLukawskiSam: I'll try to verify it.

TomaszLukawskiSam: 37 is correct because of `box-sizing: border-box;`

HunseopJeong: Oh, I see. Thank you.

HunseopJeong (approved):
LGTM

PR #993, opened 2020-04-22
(SVACE) fix major issues (part 2) 

grzegorz-czajkowski:
Contains the following commits:
 * BaseKeyboardSupport.js (SVACE): early return in case of wrong direction
 * Checkboxradio.js (SVACE): do not always return the same value
 * DropdownMenu.js (SVACE): add return to _configure
 * BaseKeyboardSupport.js (SVACE): add return to getDistanceByDirection
 * Drawer.js (SVACE): check right position
 * Dimmer.js (SVACE): add return to _refreshValue
 * BaseKeyboardSupport.js (SVACE): remove return from _onShortPress

### DIFF
--- a/src/js/core/widget/BaseKeyboardSupport.js
+++ b/src/js/core/widget/BaseKeyboardSupport.js
@@ -596,6 +596,7 @@
 						distanceRest: distances.bottomRest
 					};
 				}
+				return null;
 			}
 
 			/**
--- a/src/js/core/widget/BaseKeyboardSupport.js
+++ b/src/js/core/widget/BaseKeyboardSupport.js
@@ -1189,16 +1189,14 @@
 			prototype._onShortPress = function (event) {
 				var self = this;
 
-				if (!ns.getConfig("keyboardSupport", false)) {
-					return false;
+				if (ns.getConfig("keyboardSupport", false)) {
+					// set focus on next element
+					focusOnNeighborhood(self, self.keyboardElement || self.element, {
+						current: activeElement || getFocusedLink(),
+						event: event,
+						key: event.keyCode
+					});
 				}
-
-				// set focus on next element
-				focusOnNeighborhood(self, self.keyboardElement || self.element, {
-					current: activeElement || getFocusedLink(),
-					event: event,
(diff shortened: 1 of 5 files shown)


src/js/core/widget/BaseKeyboardSupport.js
```diff
@@ -1187,7 +1187,7 @@
 				var self = this;
 
 				if (!ns.getConfig("keyboardSupport", false)) {
```
TomaszLukawskiSam: I wondering about inversion of this condition and move the code below into this conditional.

TomaszLukawskiSam: I cannot verify this patch is solving SVACE issue. How you verify it?

grzegorz-czajkowski: I don't know the way to check if this issue is solved or not. Usually next run tells us if we achieved required number of svace issues.
I just tested UIComponets app if there is no regression after this change.

grzegorz-czajkowski: Both options are fine to me. I can change the code according to your comment.

TomaszLukawskiSam: I prefer do inversion of condition and move the code into the condition block.

I reviewed all commit in this PR. I think this is only thing to fix.

TomaszLukawskiSam (changes requested):

TomaszLukawskiSam:
I installed SonarLint in my VSC and enabled several rules enabled also in SVACE:
```
{
    "sonarlint.rules": {
        "javascript:S3801": {
            "level": "on"
        },
        "javascript:S3271": {
            "level": "on"
        },
        "javascript:S3758": {
            "level": "on"
        }
    }
}
```

obraz

src/js/core/widget/BaseKeyboardSupport.js
grzegorz-czajkowski: Done.

TomaszLukawskiSam (approved):

PR #1000, opened 2020-04-23
[mobile] Change default color of checkbox and radio button

DohyungLim:
- The default is inactive color.
- When checked, the active color is used.

### DIFF
--- a/src/css/profile/mobile/common/checkbox.less
+++ b/src/css/profile/mobile/common/checkbox.less
@@ -29,13 +29,16 @@ input[type="checkbox"].ui-checkbox {
 		width: 100%;		
 		height: 100%;
 		bottom: 0;
-		background-color: var(--control-active-color);
+		opacity: 0.8;
+		background-color: var(--control-inactive-color);
 		mask-image: url("images/3_Controllers/Checkbox/sem_checkedtextview_check_to_on_mtrl_000.svg");
 		mask-size: 100%;
 		mask-position: center;
 	}
 
 	&:checked::after {
+		opacity: 1;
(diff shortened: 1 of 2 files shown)


src/css/profile/mobile/common/radio.less
```diff
@@ -25,14 +25,15 @@ input[type="radio"].ui-radio {
 		height: 100%;
 		width: 100%;
 		bottom: 0;
-		background-color: var(--control-active-color);
+		background-color: var(--control-inactive-color);
```
singa2000, line 29: How about the opacity?

HunseopJeong: Right, Need to apply the opacity when inactive state.

src/css/profile/mobile/common/checkbox.less
```diff
@@ -29,13 +29,14 @@ input[type="checkbox"].ui-checkbox {
 		width: 100%;		
 		height: 100%;
 		bottom: 0;
-		background-color: var(--control-active-color);
+		background-color: var(--control-inactive-color);
```
HunseopJeong, line 33: ditto.

src/css/profile/mobile/common/radio.less
DohyungLim: done.

HunseopJeong: It would be good to use the @fade, but I need a check whether fade is working correctly on emulator.
I will check it later, there is no problem now.

HunseopJeong (approved):
LGTM

singa2000 (approved):
Good

PR #1001, opened 2020-04-23
UIComponents (TextInput): save data on "Save" button

grzegorz-czajkowski:
[Issue] N/A (JIRA issue)

[Problem] After click on "Save" button, nothing happens but
	expectation is that inserted data are saved and current page
	disappears. Also saved data should be filled if user navigates
	back to this page.

[Solution]
	Save data on "Save" button and navigate back. Fill the data on
	page start if they are available.
	window.sessionStorage has been used so data are available during
	session which means after application restart data are gone and
	user/tester is not confused by seeing data entered long time ago.


grzegorz-czajkowski:
Added window prefix for sessionStorage for better code readability as it was discussed internally with @TomaszLukawskiSam

lslachciak (approved):

HunseopJeong (approved):
LGTM

PR #1004, opened 2020-04-23
(UIComponents) fix for Master On-Off example

TomaszLukawskiSam:
[Issue] SAD-801
[Problem] Wrong description "Switch off disabled" when it is enabled and on
[Solution] Master On-Off example in UIComponents has been changed.
 Disabled by default On-Off switches have been removed because make logical conflict
 with Master Switch On-Off widget. That case is not described by UX guideline.


lslachciak (approved):

HunseopJeong (approved):
LGTM

singa2000:
Ok. In case of Master on/Off switch we don't consider the disabled status.

PR #1006, opened 2020-04-23
Fix for getPrefixedStyleValue method

TomaszLukawskiSam:
[Issue] SVACE-609482
[Problem] Function always returns the same value
[Solution] Method has been changed. "Return" statment was
 changed to "break" loop.

### DIFF
--- a/src/js/core/util/DOM/css.js
+++ b/src/js/core/util/DOM/css.js
@@ -350,10 +350,10 @@
 
 			/**
 			 * Returns style value for css property with browsers prefixes
-			 * @method getPrefixedStyle
+			 * @method getPrefixedStyleValue
 			 * @param {HTMLStyle} styles
 			 * @param {string} property
-			 * @return {Object}
+			 * @return {string|undefined}
 			 * @member ns.util.DOM
 			 * @static
 			 */
@@ -366,7 +366,7 @@
 					if (prefixedProperties.hasOwnProperty(key)) {
 						value = styles[prefixedProperties[key]];
 						if (value && value !== "none") {
-							return value;
+							break;
 						}
 					}
 				}


src/js/core/util/DOM/css.js
```diff
@@ -366,7 +366,7 @@
 					if (prefixedProperties.hasOwnProperty(key)) {
 						value = styles[prefixedProperties[key]];
 						if (value && value !== "none") {
-							return value;
+							break;
```
HunseopJeong, line 369: I don't understand your comment.
'getPrefixedStyleValue' finds until there is a valid value and returns the value.
It doesn't seem to always return the same value.

HunseopJeong: @method name is wrong. need to fix.
Plus, @return seems a string.

grzegorz-czajkowski: I also don't understand this issue but replacing return with break may not solve this SVACE issue. Maybe loop always iterates over the same key/value?

TomaszLukawskiSam: > I don't understand your comment.
> 'getPrefixedStyleValue' finds until there is a valid value and returns the value.
> It doesn't seem to always return the same value.

This is issue reported by SVACE. For me, the previous version of code is clear, but Sonar Linter found a problem with return.

TomaszLukawskiSam: > I also don't understand this issue but replacing return with break may not solve this SVACE issue. Maybe loop always iterates over the same key/value?

I installed SonarLinter localy in my VSC. The issue not occurred after this change. The `prefixedProperties` is object like this:
```
{
	webkit: "-webkit-" + value,
	moz: "-moz-" + value,
	o: "-ms-" + value,
	ms: "-o-" + value,
	normal: value
}
```
Method check in loop if value like ` "-webkit-" + value` exists in `style` object. If exists then return this value.

HunseopJeong: Is it okay to return the 'value' not 'break'?

grzegorz-czajkowski: To me, previous version was clear but behavior is not changed with proposed change and should not introduce regression. Let's satisfy SVACE :)

grzegorz-czajkowski (approved):
Looks good to me.

HunseopJeong (approved):
LGTM

PR #1007, opened 2020-04-23
Fix for browser.ie detect method

TomaszLukawskiSam:
[Issue] SVACE 609481
[Problem] Method browser.ie must return the same type
[Solution] Method returns boolean or number type
 but number result is used only in the same module.
 After this change method always returns boolean type
 according to jsDoc

### DIFF
--- a/src/js/core/support.js
+++ b/src/js/core/support.js
@@ -54,7 +54,8 @@
 				opera = window.opera,
 				operamini = window.operamini && ({}).toString.call(window.operamini) === "[object OperaMini]",
 				blackBerry,
-				testDivStyle = testDiv.style;
+				testDivStyle = testDiv.style,
+				ieVersion;
 
 			testDiv.id = "jquery-mediatest";
 			fakeBody.appendChild(testDiv);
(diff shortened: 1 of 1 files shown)


src/js/core/support.js
```diff
@@ -222,6 +223,17 @@
 				return div.getBoundingClientRect !== undefined;
 			}
 
+			ieVersion = (function getIEVersion() {
```
grzegorz-czajkowski, line 226: Could this be a member of browser object?
```
browser {
  isIE: func()
  ieVersion: func()
}
```

TomaszLukawskiSam: I don't want to change API.
Now, `ieVersion` is called once.
Version of IE is compared only in one place. `this.browser.ie >= 7`

grzegorz-czajkowski: How about just adding ieVersion to browser object instead of defining it separately? It looks wired to have ieVersion() function where we have browser object with ie property.

TomaszLukawskiSam: As I wrote before I don't want to change API. I just try to make refactoring for fix SVACE issue.

grzegorz-czajkowski (approved):
Ok, let's satisfy SAVE

PR #1009, opened 2020-04-24
List (mobile): Add a height between title and sub text

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/980
[Problem] There is no margin when the sub text is placed upper title text
[Solution] Add a height to match with GUI feedback

### DIFF
--- a/src/css/profile/mobile/common/listview.less
+++ b/src/css/profile/mobile/common/listview.less
@@ -173,6 +173,13 @@ tau-expandable {
 				color: var(--text-secondary-color);
 				line-height: 15 * @px_base;
 
+				// line height between title and sub
+				&+.ui-li-text-title::before {
+					content: "";
+					display: block;
+					height: 4 * @px_base;
+				}
+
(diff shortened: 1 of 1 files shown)


src/css/profile/mobile/common/listview.less
```diff
@@ -173,6 +173,13 @@ tau-expandable {
 				color: var(--text-secondary-color);
 				line-height: 15 * @px_base;
 
+				// line height between title and sub
+				&+.ui-li-text-title::before {
+					content: "";
+					display: block;
+					height: 4 * @px_base;
+				}
+
```
singa2000, line 182: You already set the "ui-li-text-sub::before" by "height: 4 * @px_base;"
Is it also needed?

				// line height between title and sub
				&+.ui-li-text-sub::before {
					content: "";
					display: block;
					height: 4 * @px_base;
				}

singa2000 (approved):
LGTM

PR #1010, opened 2020-04-24
UIComponents (OnOffSwitch) : fix Back key operation

singa2000:
[Issue] SAD-800
[Problem] Back key button is not working
[Solution] Add back key  operation


HunseopJeong (approved):
LGTM

PR #1020, opened 2020-04-24
UIComponents (NumberPicker): save number on "Set" button

grzegorz-czajkowski:
[Issue] SAD-798
[Problem] After click on button "Set", current number not saved
for Simple and Accelerated Number Pickers.
[Solution] Save the number to sessionStorage and restore it on "pageshow".


TomaszLukawskiSam (approved):
ok

PR #1023, opened 2020-04-27
[mobile] Show divider dynamically when list scrolls

DohyungLim:
- In the case of scrolling lists, dividers are created.
- The default is only the bottom divider.
- Scrolling shows dividers at the top and bottom.
- When scrolling down to the end, the divider is visible only at the top.

### DIFF
--- a/src/css/profile/mobile/common/popup.less
+++ b/src/css/profile/mobile/common/popup.less
@@ -54,6 +54,28 @@
 		white-space: nowrap;
 		.font(medium);
 
+		&::after {
+			content: "";
+			position: absolute;
+			left: 24 * @px_base;
+			width: calc(~"100%" - 48 * @px_base);
+			height: 0.75 * @px_base;
+			stroke-width: 0.5;
+			visibility: hidden;
+			background-color: var(--popup-scroll-divider-color);
+
+			@media (orientation: portrait) {
+				top: 68 * @px_base;
+			}
+			@media (orientation: landscape) {
+				top: 45 * @px_base;
+			}
+		}
+
+		&.topDivider::after {
+			visibility: visible;
+		}
+
 		&.ui-popup-header-has-subtitle {
 			padding: 19.5 * @unit_base 32 * @unit_base;
 			line-height: 57 * @unit_base;
@@ -139,10 +161,12 @@
 		}
 
 		@media (orientation: portrait) {
-			padding: 0 * @px_base 24 * @px_base 24 * @px_base 24 * @px_base;
+			padding: 0 24 * @px_base 0 24 * @px_base;
+			margin-bottom: 24 * @px_base;
 		}
 		@media (orientation: landscape) {
-			padding: 0 * @px_base 24 * @px_base 11 * @px_base 24 * @px_base;
+			padding: 0 24 * @px_base 0 24 * @px_base;
+			margin-bottom: 11 * @px_base;
 		}
 
 		&.ui-popup-has-time-picker, &.ui-popup-has-date-picker {
@@ -162,6 +186,28 @@
 		flex-shrink: 0;
 		flex-grow: 0;
 		width: 100%;
+
+		&::before {
+			content: "";
+			position: absolute;
+			left: 24 * @px_base;
+			width: calc(~"100%" - 48 * @px_base);
+			height: 0.75 * @px_base;
+			stroke-width: 0.5;
+			visibility: hidden;
+			background-color: var(--popup-scroll-divider-color);
+
+			@media (orientation: portrait) {
+				bottom: 80 * @px_base;
+			}
+			@media (orientation: landscape) {
+				bottom: 50 * @px_base;
+			}
+		}
+
+		&.bottomDivider::before {
+			visibility: visible;
+		}
 
 		.ui-btn {
 			height: 36 * @px_base;
--- a/src/js/core/widget/core/Popup.js
+++ b/src/js/core/widget/core/Popup.js
@@ -630,7 +630,33 @@
 					// set state of popup
 					self.state = states.CLOSED;
 				}
+
+				if (self._ui.content.scrollHeight > self._ui.content.clientHeight) {
+					self._ui.footer.classList.add("bottomDivider");
+				}
 			};
+
+			/**
+			 * Scroll event
+			 * @method _onScroll
+			 * @protected
+			 * @member ns.widget.core.Popup
+			 */
+			prototype._onScroll = function () {
+				var self = this,
+					content = self._ui.content;
+
+				if (content.scrollTop === 0) {
+					self._ui.header.classList.remove("topDivider");
+					self._ui.footer.classList.add("bottomDivider");
+				} else if (content.scrollHeight - content.clientHeight === content.scrollTop) {
+					self._ui.header.classList.add("topDivider");
+					self._ui.footer.classList.remove("bottomDivider");
+				} else {
+					self._ui.header.classList.add("topDivider");
+					self._ui.footer.classList.add("bottomDivider");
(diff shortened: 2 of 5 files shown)


src/js/core/widget/core/Popup.js
```diff
@@ -644,6 +648,23 @@
 				eventUtils.on(self._ui.page, "pagebeforehide", self, false);
 				eventUtils.on(window, "resize", self, false);
 				eventUtils.on(document, "vclick", self, false);
+
+				self._ui.content.addEventListener("scroll", function () {
```
TomaszLukawskiSam: Please exclude this anonymous function to named function, eg. `prototype._onScroll = function() { `

TomaszLukawskiSam: Second thing,
remove this listener in `_unbindEvents` method

src/js/core/widget/core/Popup.js
```diff
@@ -644,6 +648,23 @@
...
+					eventUtils.trigger(self._ui.content, "scrollstart");
```
TomaszLukawskiSam: property `self._ui.content` repeated several times. You can do local variable `var content = self._ui.content`

src/css/profile/mobile/common/popup.less
```diff
@@ -163,6 +187,28 @@
 		flex-grow: 0;
 		width: 100%;
 
+		&::before {
```
TomaszLukawskiSam, line 190: In header you used `::after` but in footer you use `::before`. Is possible to use the same type of pseudo element?

src/css/profile/mobile/common/popup.less
```diff
@@ -163,6 +187,28 @@
...
+			@media (orientation: portrait) {
+				bottom: 80 * @px_base;
+			}
+			@media (orientation: landscape) {
+				bottom: 50 * @px_base;
+			}
+		}
+
+		&.bottomDivider::before {
+			visibility: visible;
```
TomaszLukawskiSam, line 209: Element hidden by visible still hold a place. Is this intended?

TomaszLukawskiSam (changes requested):

src/js/core/widget/core/Popup.js
DohyungLim: done.

src/css/profile/mobile/common/popup.less
DohyungLim: A divider should be created at the bottom of the header.
That's why I use ::after.
The divider is also created in the upper part of the footer, so ::before should be used.
This is the intended result.

src/css/profile/mobile/common/popup.less
DohyungLim: That's right. 
The only place in the dialog where scrolls are generated is in cases where there are many list items.
I intentionally applied the visibility option.

src/js/core/widget/core/Popup.js
```diff
@@ -630,8 +630,61 @@
 					// set state of popup
 					self.state = states.CLOSED;
 				}
+
+				if (self._ui.content.scrollHeight > self._ui.content.clientHeight) {
+					self._ui.footer.classList.add("bottomDivider");
+				}
 			};
 
+			prototype._onScroll = function (content) {
```
TomaszLukawskiSam: You have twice the same method

src/js/core/widget/core/Popup.js
```diff
@@ -630,8 +630,61 @@
...
+			}
+
+			/**
+			 * Scroll event
+			 * @method _onScroll
+			 * @param {Element} content
+			 * @protected
+			 * @member ns.widget.core.Popup
+			 */
+			prototype._onScroll = function (content) {
```
TomaszLukawskiSam: The same method above.

TomaszLukawskiSam: This method is wrong. You should add listener in _bindEvents

src/js/core/widget/core/Popup.js
```diff
@@ -644,6 +697,8 @@
 				eventUtils.on(self._ui.page, "pagebeforehide", self, false);
 				eventUtils.on(window, "resize", self, false);
 				eventUtils.on(document, "vclick", self, false);
+
```
TomaszLukawskiSam: in this place:
`eventUtils.on(self._ui.content, "scroll", self, false);`

src/js/core/widget/core/Popup.js
```diff
@@ -659,6 +714,8 @@
 				eventUtils.off(self._ui.page, "pagebeforehide", self, false);
 				eventUtils.off(window, "resize", self, false);
 				eventUtils.off(document, "vclick", self, false);
+
+				self._onScroll(null);
```
TomaszLukawskiSam: instead of this line,
`eventUtils.off(self._ui.content, "scroll", self, false);`

src/js/core/widget/core/Popup.js
```diff
@@ -644,6 +697,8 @@
 				eventUtils.on(self._ui.page, "pagebeforehide", self, false);
 				eventUtils.on(window, "resize", self, false);
 				eventUtils.on(document, "vclick", self, false);
+
+				self._onScroll(self._ui.content);
```
TomaszLukawskiSam: move this method to `prototype.handleEvent`
```
					case "scroll":
						self._onScroll(event);
						break;
```

src/js/core/widget/core/Popup.js
DohyungLim: done.

src/js/core/widget/core/Popup.js
DohyungLim: done.

src/js/core/widget/core/Popup.js
DohyungLim: sorry. I will remove one function.

done.

src/js/core/widget/core/Popup.js
DohyungLim: done.

src/js/core/widget/core/Popup.js
DohyungLim: done.

src/js/core/widget/core/Popup.js
```diff
@@ -630,8 +630,40 @@
...
+			 * Scroll event
+			 * @method _onScroll
+			 * @protected
+			 * @member ns.widget.core.Popup
+			 */
+			prototype._onScroll = function () {
+				var self = this,
+					content = self._ui.content;
+
+				content.addEventListener("scroll", function () {
```
TomaszLukawskiSam: You don't need this line.

src/js/core/widget/core/Popup.js
```diff
@@ -630,8 +630,40 @@
...
+					} else if (content.scrollHeight - content.clientHeight === content.scrollTop) {
+						self._ui.header.classList.add("topDivider");
+						self._ui.footer.classList.remove("bottomDivider");
+					} else {
+						self._ui.header.classList.add("topDivider");
+						self._ui.footer.classList.add("bottomDivider");
+					}
+
+					eventUtils.trigger(content, "scrollupdate");
+				}, false);
```
TomaszLukawskiSam: and this line.

src/js/core/widget/core/Popup.js
DohyungLim: done.

src/js/core/widget/core/Popup.js
DohyungLim: done.

TomaszLukawskiSam (approved):
Looks good to me

PR #1024, opened 2020-04-27
(svace) Replacing localstorege by document.cookie in PanelChanger

TomaszLukawskiSam:
[Issue] SVACE-609506
[Problem] S3271 SonarJS vulnerability issue
[Solution] Feature of localstorage has been replaced by document.cookie

### DIFF
--- a/src/js/core/widget/core/PanelChanger.js
+++ b/src/js/core/widget/core/PanelChanger.js
@@ -37,6 +37,7 @@
 			"../../event",
 			"../../util/selectors",
 			"../../util/object",
+			"../../util/cookie",
 			"../../history/manager",
 			"../BaseWidget",
 			"../core",
@@ -48,6 +49,7 @@
 			var BaseWidget = ns.widget.BaseWidget,
 				selectors = ns.util.selectors,
 				object = ns.util.object,
+				cookie = ns.util.cookie,
 				engine = ns.engine,
 				page = ns.widget.core.Page,
 				panel = ns.widget.core.Panel,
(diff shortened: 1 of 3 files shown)


src/js/core/widget/core/PanelChanger.js
```diff
@@ -52,6 +52,7 @@
 				page = ns.widget.core.Page,
 				panel = ns.widget.core.Panel,
 				events = ns.event,
+				COOKIE_MAX_AGE = 24 * 60 * 60, // one day;
```
grzegorz-czajkowski: Wouldn't be user surprised that his values gone after one day?. There is no expiration date for localStorage. This changes behavior a bit.

src/js/core/widget/core/PanelChanger.js
```diff
@@ -83,6 +84,28 @@
 			PanelChanger.classes = classes;
 			PanelChanger.prototype = prototype;
 
+			function writeToCookie(storageName, value) {
```
lslachciak: I think this is function which can be useful in future. How about moving it to utils?

grzegorz-czajkowski: Those functions looks like good candidates for utils since they could be used by different modules. What do you think?

TomaszLukawskiSam: I'll move these methods to `tau.util.cookie`

TomaszLukawskiSam: Ok. As above, I move these functions to 'tau.util.cookie'

src/js/core/widget/core/PanelChanger.js
TomaszLukawskiSam: This arbitrary value but seems unnecessary in this case. The cookie should expired with session.

TomaszLukawskiSam:
@lslachciak @grzegorz-czajkowski 
Hi, I changed my PR. Could I ask you for review again?

grzegorz-czajkowski:
@TomaszLukawskiSam  Please check CI

grzegorz-czajkowski (approved):
Looks fine.

Issue #1025, opened 2020-04-27
extended appbar should not be applied for height up to 579 


lmslachciak:
see GUI Guide ver 1.8 p.43
GUI feedback p. 3

TomaszLukawskiSam:
@lslachciak 
Patch has been merged. Can we close the issue?

lslachciak:
works well, closing issue

PR #1049, opened 2020-05-04
AppBar: disallow expand for screen below height 579

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1025
[Problem] extended appbar should not be applied for height up to 579
[Solution]
 - added new method .expand(), .collapse()
 - added listeners for resize
 - added refresh page content
 - added options for disable animation and disable expanding

[Test use case]
- open UIComponents > Lists > List Content area
- Expand AppBar by drag
- change device orientation to landscape
Result: The AppBar should be collapsed.

### DIFF
--- a/src/js/core/widget/core/Appbar.js
+++ b/src/js/core/widget/core/Appbar.js
@@ -46,6 +46,7 @@
 					COLLAPSED: "COLLAPSED",
 					DRAGGING: "DRAGGING"
 				},
+				SCREEN_HEIGHT_LIMIT_FOR_EXPANDING = 579,
 				Appbar = function () {
 					var self = this;
 
@@ -57,7 +58,7 @@
 						page: null,
 						selectAll: null,
 						bottomBar: null
-					},
+					};
 					self._appbarState = states.COLLAPSED;
 					self._dragStartingHeight = 0;
 					self._currentHeight = 0;
@@ -78,6 +79,7 @@
 					dragging: classPrefix + "-dragging",
 					controlsContainer: classPrefix + "-controls-container",
 					expandedTitleContainer: classPrefix + "-expanded-title-container",
+					animationFast: classPrefix + "-animation-fast",
 					selectAll: "ui-label-select-all",
 					bottomBar: "ui-bottom-bar",
 					hidden: "ui-hidden"
@@ -109,7 +111,8 @@
 				},
 				defaults = {
 					titleType: "singleLine", // "multiline", "subtitle"
-					expandingEnabled: true
+					expandingEnabled: true,
+					animation: true
 				};
 
 			Appbar.prototype = prototype;
@@ -121,7 +124,10 @@
 
 				self._initExpandedContainer(element);
 
+				self._setAnimation(element, self.options.animation);
 				self._appbarState = states.COLLAPSED;
+				self._validateExpanding();
+
 				self._ui.page = utilSelectors.getClosestBySelector(element, Page.selector);
 				self._ui.selectAll = element.querySelector("." + classes.selectAll + " input[type='checkbox']");
 				self._ui.bottomBar = self._ui.page.querySelector("." + classes.bottomBar);
@@ -253,6 +259,7 @@
 					self._ui.page,
 					"scrollboundary drag dragstart dragend scrollstart change pagebeforeshow popupshow popuphide",
 					self);
+				window.addEventListener("resize", self, false);
 			};
 
 			/**
@@ -292,6 +299,9 @@
 						break;
 					case "popuphide":
 						self._onPopupHide();
+						break;
+					case "resize":
+						self._onResize();
 						break;
 				}
 			};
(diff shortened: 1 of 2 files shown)


src/js/core/widget/core/Appbar.js
```diff
@@ -351,6 +368,51 @@
 				expandedTitle.style.opacity = expandLevel;
 			};
 
+			/**
+			 * Expand AppBar
+			 * @method expand
+			 * @member ns.widget.core.Appbar
```
grzegorz-czajkowski, line 416: `@public`

src/js/core/widget/core/Appbar.js
```diff
@@ -351,6 +368,51 @@
...
+			 */
+			prototype.expand = function () {
+				var self = this;
```
grzegorz-czajkowski: How about defining element as  well, we are going to use it twice.

```
var self = this,
    element = self.element;
```

src/js/core/widget/core/Appbar.js
```diff
@@ -351,6 +368,51 @@
...
+			};
+
+			/**
+			 * Animation transition end event handler
+			 * @method _onTransitionEnd
+			 * @member ns.widget.core.Appbar
+			 * @protected
+			 */
+			prototype._onTransitionEnd = function () {
+				ns.engine.getBinding(this._ui.page, "Page").refresh();
```
grzegorz-czajkowski, line 443: I think it's worth adding comment why the page needs to be refreshed.

src/js/core/widget/core/Appbar.js
```diff
@@ -351,6 +368,51 @@
...
+			};
+
+			/**
+			 * Fold AppBar
+			 * @method fold
+			 * @member ns.widget.core.Appbar
```
grzegorz-czajkowski, line 449: `@public`

src/js/core/widget/core/Appbar.js
```diff
@@ -351,6 +368,51 @@
...
+			 */
+			prototype.fold = function () {
```
grzegorz-czajkowski: Consider the name collapse as it often can be see together with expand feature.

src/js/core/widget/core/Appbar.js
```diff
@@ -413,6 +468,33 @@
 				}
 			};
 
+			/**
+			 * Verification of screen height
+			 * @method _verifyExpanding
+			 * @member ns.widget.core.Appbar
+			 * @protected
+			 */
+			prototype._verifyExpanding = function () {
```
grzegorz-czajkowski: The code logic suggests "_setExpanding" name.

src/js/core/widget/core/Appbar.js
```diff
@@ -584,6 +667,38 @@
...
 
+			/**
+			 * Set expanding option
+			 * @method _setExpandingEnabled
+			 * @param {HTMLElement} element
+			 * @param {boolean} enabled
+			 * @member ns.widget.core.Appbar
+			 * @protected
+			 */
+			prototype._setExpandingEnabled = function (element, enabled) {
```
grzegorz-czajkowski, line 731: Not used. Please remove it.

grzegorz-czajkowski (changes requested):

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: this is handler for widget.option("expandingEnabled", boolean)

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: @grzegorz-czajkowski @lslachciak 
We didn't use `@public` (BaseWidget.js) but I see that during of project's live `@public` tag began to appear in source code (wearable/Selector.js). I don't know whether to keep this rule or not

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: Thanks, I'll change this.

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: I wondered about it before. The "_set" is not also good because it is not the "set" we don't have parameter of function ... What you think about "_validate"

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: You are right

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/Appbar.js
TomaszLukawskiSam: Done.

src/js/core/widget/core/Appbar.js
lslachciak: I think we should use '@public' tag. This will be a good way to show which API can be used by developers

grzegorz-czajkowski (approved):
LGTM.

DohyungLim (approved):
LGTM.

Issue #1051, opened 2020-05-05
[wearable] Presentation Views -> List -> Main + SubText


lmslachciak:

grzegorz-czajkowski:
Patch merged. Issue can be closed.

lslachciak:
In order to access Main + Sub text example you need to go
Presentation View --> List --> **SubText** --> **SubText**
I think we shouldn't duplicate Subtext menu.
How about 
Presentation View --> List --> **Main** --> **SubText**

grzegorz-czajkowski:
SubText has 4 examples:
 - Sub text
 - Sub text (editable)
 - Sub text max (fade on right) https://github.com/Samsung/TAU/issues/1052
 - Sub text max (fade on left) https://github.com/Samsung/TAU/issues/1052

That's why I created separated menu for future implementation, also added comment in the code for this 
https://github.com/Samsung/TAU/blob/c322e506039bac5544059cd3113ee1a29fb191ce/examples/wearable/UIComponents/contents/presentation-views/list/sub-text/index.html#L32

lslachciak:
I'll prepare PR with new concept of menus

PR #1065, opened 2020-05-06
List Sub text (Wearable): restore example and update style

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/1051
[Problem] Lack of example
[Solution] Restore and update example from [1] and update
	TAU style for Sub text.

[1] examples/wearable/UIComponents/contents/etc/list/subtext/list_subtext_1line.html


korneliakobiela (approved):
LGTM

Issue #1067, opened 2020-05-07
DropDown: More menu goes out of screen


TomaszLukawskiSam:
![obraz](https://user-images.githubusercontent.com/29534410/81265855-0d16df00-9044-11ea-8752-33d3f74157aa.png)

Fix in PR https://github.com/Samsung/TAU/pull/1048 makes issue.

[Test]
UIComponents > AppBar > More options list (Max width)

TomaszLukawskiSam:
Patch has already been merged.

PR #1069, opened 2020-05-07
DropDown: fix left position depending on screen width

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1067
[Problem] More menu goes out of screen
[Solution] Position of DropDown menu has to takes account
 screen size.

### DIFF
--- a/src/js/profile/mobile/widget/DropdownMenu.js
+++ b/src/js/profile/mobile/widget/DropdownMenu.js
@@ -1024,7 +1024,10 @@
 					ui = self._ui,
 					optionHeight = ui.elOptionContainer.offsetHeight,
 					listItemWidthOffsets = [].slice.call(ui.elOptionContainer.children).map(mapItemWidth),
-					biggestListItemWidth = Math.max.apply(Math, listItemWidthOffsets),
+					optionContainerStyle = window.getComputedStyle(ui.elOptionContainer),
+					biggestListItemWidth = Math.max.apply(Math, listItemWidthOffsets) +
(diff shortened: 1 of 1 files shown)


grzegorz-czajkowski:
@TomaszLukawskiSam I checkted latest TAU code and can not reproduce this issue using steeps mentioned in https://github.com/Samsung/TAU/issues/1067. Did you set specific width? I can see similar issue while window resizing. Could you specify steeps to reproduce?

grzegorz-czajkowski:
Able to reproduce this on Tizen Emulator.

src/js/profile/mobile/widget/DropdownMenu.js
```diff
@@ -1076,7 +1084,10 @@
 					ui.elOptionWrapper.classList.add(classes.verticalMargins);
 				}
 
-				optionStyle += "top: " + offsetTop + "px; width: " + width + "px; max-height: " + height + "px;";
+				optionStyle += "top: " + offsetTop + "px; max-height: " + height + "px;";
+				if (options.inline === true) {
```
grzegorz-czajkowski: Not sure if this option is still supported and works as it should be. Are we sure that width should be excluded  one inline option is specified?

TomaszLukawskiSam: You are right, `inline` change is out of scope this issue.

grzegorz-czajkowski (approved):
Looks good to me. Verified.

Issue #1074, opened 2020-05-08
Dialogs --> Date & time picker (12h)


lmslachciak:
Please refer to:
GUI Guide p. 138 (v.1.9)
Components UI p. v.17 rel. 2 (p. 94-100)
Reference_SR_200306 (p. 3-6)

singa2000:
F.Y.I.
The Date&Time Picker can use both with dialog and component itself used with body.

In case of Reference app, it use by component itself used with body.

lslachciak:
Right. at this moment the most important is usage in body.
I updated issue description with appropriate reference

TomaszLukawskiSam:
@lslachciak 
I think the issue can be close.

lslachciak:
works fine, although two issues exist
https://github.com/Samsung/TAU/issues/1104
https://github.com/Samsung/TAU/issues/1108

PR #1082, opened 2020-05-11
DateTimePicker: widget implementation and sample

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1074
[Problem] DateTime picker is not implemented yet
[Solution]
  - added new DataTimePicker widget
  - widget triggers event "selected" on date click/tap
  - widget accepts "data-value" attribute to set value
    in the same way like Date constructor, 
    example: data-value="December 17, 1995 03:24:00"
  - added examples

[Known issues]
 - lock scroll during Spin drag

[Demo]
obraz

### DIFF
--- a/examples/mobile/UIComponents/components/dialogs/picker.html
+++ b/examples/mobile/UIComponents/components/dialogs/picker.html
@@ -43,6 +43,11 @@
 						Date picker (Wheel view)
 					</a>
 				</li>
+				<li class="ui-li-anchor">
+					<a data-position-to="window" data-rel="popup" href="date-time-picker-wheel">
+						Date Time picker (Wheel view)
+					</a>
+				</li>
 			</ul>
 			<div class="ui-popup" id="date-picker-calendar-dialog">
 				<div class="ui-popup-header">
@@ -86,6 +91,16 @@
 			<div class="ui-popup" id="date-picker-wheel">
 				<div class="ui-popup-content ui-popup-has-date-picker">
 					<div class="ui-date-picker" data-view="wheel"></div>
+				</div>
+				<div class="ui-popup-footer">
+					<button class="ui-btn" data-style="flat">Cancel</button>
+					<div class="ui-li-divider"></div>
+					<button class="ui-btn" data-style="flat">Done</button>
+				</div>
+			</div>
+			<div class="ui-popup" id="date-time-picker-wheel">
+				<div class="ui-popup-content ui-popup-has-date-picker">
+					<div class="ui-datetime-picker" data-view="wheel" data-format="12" data-value="2020-03-12 14:46:20"></div>
 				</div>
 				<div class="ui-popup-footer">
 					<button class="ui-btn" data-style="flat">Cancel</button>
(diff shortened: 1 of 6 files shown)


examples/mobile/UIComponents/components/dialogs/picker.html
```diff
@@ -93,6 +98,16 @@ <h1>
 					<button class="ui-btn" data-style="flat">Done</button>
 				</div>
 			</div>
+			<div class="ui-popup" id="date-time-picker-wheel">
+				<div class="ui-popup-content ui-popup-has-date-picker">
+					<div class="ui-datetime-picker" data-view="wheel" data-format="12" data-value="2020-03-12"></div>
```
singa2000: Does only set the date through "data-value"? can't set the time?
And just question, if we can set the date, the format is 2020-03-12 not like 2020:03:12?

TomaszLukawskiSam: Thank you for review. Work is in progress now, but you are right. I'll do it for accept the same string value as `Data` constructor.
```
let birthday = new Date('December 17, 1995 03:24:00')
let birthday = new Date('1995-12-17T03:24:00')
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

or better the API should be the same as:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

TomaszLukawskiSam:
@lslachciak 
I implemented DateTime Picker to ScheduleApp but I found issue with GUI guideline.
According to GUI guideline, the DateTime Picker has top/bottom padding (p. 140 v.2.0). Is this proper?

obraz

lslachciak:
This padding from GUI guideline doesn't look good in sample app. Let's resign form it now, and clarify with it with UX team. @singa2000 could you help us in this matter?

src/css/profile/mobile/common/datetimepicker.less
```diff
@@ -0,0 +1,105 @@
+.ui-datetime-picker {
+	display: flex;
+	height: 164 * @px_base;
+	margin-left: 4%;
+	margin-right: 6%;
+	padding-top: 60 * @px_base;
```
lslachciak, line 6: let's comment out this until clarified with UX

TomaszLukawskiSam: Thanks.

TomaszLukawskiSam: App sfter remove padding.
obraz

singa2000:
Sure. I'll check it with UX team.

singa2000:
I check it with UX team, it's right  GUI Guide 141.
Top padding is 60, bottom padding is 81.

src/css/profile/mobile/common/datetimepicker.less
TomaszLukawskiSam: We have answer from UX team. The padding remain.

HunseopJeong (approved):
LGTM

Issue #1097, opened 2020-05-15
List divider bottom round corner


TomaszLukawskiSam:
The bottom of list is not rounded when it's scrolling.

![obraz](https://user-images.githubusercontent.com/29534410/82039030-abd6b780-96a4-11ea-8f65-22858173772c.png)

TomaszLukawskiSam:
Patch has already been merged. The issue can be close.

lslachciak:
verified, works fine

Issue #1086, opened 2020-05-12
[mobile] [issue] Extending Appbar cause list area to diminish


lmslachciak:
Steps:
1. open example
http://localhost:8080/examples/mobile/SchedulingApp/create-event.html
2. extend appbar
3. collapse appbar
4. extend appbar
5. collapse appbar
Observe:
![image](https://user-images.githubusercontent.com/2246331/81698845-d6d3c800-9466-11ea-9e1b-c2b36dd04698.png)

TomaszLukawskiSam:
@lslachciak 
We can join issue with https://github.com/Samsung/TAU/issues/1097 .
I am in the process of solving the problem #1097

TomaszLukawskiSam:
The issue is solving by patch for another issue (https://github.com/Samsung/TAU/pull/1098)

TomaszLukawskiSam:
Patch has already been merged. The issue can be close.

lslachciak:
thank you @TomaszLukawskiSam issue no longer exist now.

PR #1098, opened 2020-05-18
Page: Bottom round corners

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1097 https://github.com/Samsung/TAU/issues/1086
[Problem] List divider bottom round corner
[Solution]
     - added flex style to page,
     - modified flex value for content of page to occupies all available space

### DIFF
--- a/src/js/core/widget/core/Appbar.js
+++ b/src/js/core/widget/core/Appbar.js
@@ -59,6 +59,7 @@
 						selectAll: null,
 						bottomBar: null
 					};
+					self._expandedHeight = nominalHeights.EXPANDED;
 					self._appbarState = states.COLLAPSED;
 					self._dragStartingHeight = 0;
 					self._currentHeight = 0;
@@ -86,7 +87,7 @@
 				},
 				nominalHeights = {
 					COLLAPSED: 56,
-					expanded: 56
+					EXPANDED: 56
 				},
 				containersProperties = {
 					leftIconsContainer: {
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -1421,6 +1421,11 @@ module.exports = function (grunt) {
 							"src": "src/css/profile/mobile/theme-changeable/images/3_Controllers/Radio/",
 							"dest": "dist/mobile/theme/changeable/images/3_Controllers/Radio/sprites/",
 							"fileMask": "_[0-9]+\\.svg$"
+						},
+						{
+							"src": "src/css/profile/mobile/theme-changeable/images/0_Round_corner/",
```
singa2000: For implement the round corner, if we use the both svg and shape(css), the corner line is not clear.
So as much as possible, we want to implement it without the svg, just css shape.
If there is not needed svg  file for corner line, UX guide will delete the method of using svg in UX Guide.

TomaszLukawskiSam: If we cannot use svg file as mask in footer then the only way is add "flex" style to page and page-content.
Height of "ui-content" has to strict even when AppBar will be during animation of expand. This guarantees that bottom corners will be always visible.

src/js/core/widget/core/Appbar.js
```diff
@@ -440,7 +441,7 @@
 				 * so Page should be refreshed to better fit
 				 * to new available space.
 				 */
-				ns.engine.getBinding(this._ui.page, "Page").refresh();
+				utilsEvents.trigger(this.element, "appbartransitionend");
```
singa2000, line 448: Where is defined "appbartransitionend"?
I can't find it.

TomaszLukawskiSam: This is signal to developer that AppBar has changed own size. Developer can resize height of content if it is needed. I don't want to force page refresh because this is not always required.

src/css/profile/mobile/common/maintab.less
```diff
@@ -11,6 +11,7 @@
 	background-color: var(--bottom-bar-color);
 	position: fixed;
 	bottom: 0;
+	z-index: 1000;
```
DohyungLim: Is there any reason to set the value to 1000?
Wouldn't the other cases overlap?

src/js/core/widget/core/Appbar.js
```diff
@@ -507,15 +508,17 @@
 					totalHeight = min(totalHeight, nominalHeights.expanded);
```
DohyungLim: It's not something you've modified, but I think it's good to use capital letters.

DohyungLim (commented):
2 comment.

lslachciak:
At this moment works fine, but please do rebase because I see svg dependencies

src/css/profile/mobile/common/maintab.less
TomaszLukawskiSam: I'll remove it.

lslachciak (approved):

PR #1100, opened 2020-05-18
Remove the svg file from the mask-border

HunseopJeong:
[Issue] N/A
[Problem] Radius effect is applied redundantly
[Solution] Remove mask-border among duplicated effects

### DIFF
--- a/src/css/profile/mobile/common/core.less
+++ b/src/css/profile/mobile/common/core.less
@@ -134,8 +134,6 @@
 	overflow-y: visible;
 	overflow-x: hidden;
 	flex-shrink: 1;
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam (approved):
OK

src/css/profile/mobile/common/core.less
```diff
@@ -151,8 +149,6 @@
 
 	.ui-content-area {
 		background-color: var(--background-area-color);
-		mask-border: url(images/0_Round_corner/round.svg) 77;
```
TomaszLukawskiSam, line 154: we can remove the svg file too.

singa2000 (approved):
It's good working

Issue #1104, opened 2020-05-19
DateTimePicker: lock scroll event during drag event on picker


TomaszLukawskiSam:
related to known issue: https://github.com/Samsung/TAU/pull/1082

TomaszLukawskiSam:
Patch has been merged. The issue can be close.

lslachciak:
works fine

PR #1109, opened 2020-05-20
Spin: lock scroll before drag event on widget

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1104
[Problem] During drag on Spin widget the page content is scrolling
[Solution]
 - style overflow is set to hidden on scrollable parent
  before drag event on Spin widget

### DIFF
--- a/src/js/core/widget/core/Spin.js
+++ b/src/js/core/widget/core/Spin.js
@@ -35,7 +35,8 @@
 		"../BaseWidget",
 		"../../../core/engine",
 		"../../../core/util/animation/animation",
-		"../../../core/event/gesture"
+		"../../../core/event/gesture",
+		"../../../core/util/selectors"
 	],
 	function () {
 		//>>excludeEnd("tauBuildExclude");
@@ -44,6 +45,7 @@
 			engine = ns.engine,
 			utilsEvents = ns.event,
 			gesture = utilsEvents.gesture,
+			utilSelectors = ns.util.selectors,
 
 			Animation = ns.util.Animate,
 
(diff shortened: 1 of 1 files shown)


lslachciak (commented):
this works when apopbar is not extended.
but when appabr is extended double scrolling occurs,
we may address this in separate issue if you want
?

src/js/core/widget/core/Spin.js
```diff
@@ -107,9 +111,12 @@
...
 					dragTarget: "document" // "document" | "self"
 				};
-				this._ui = {};
-				this.length = this.options.max - this.options.min + 1;
-				this._prevValue = null; // this property has to be "null" on start
+				self._ui = {
+					scrollableParent: null
+				};
+				self.length = self.options.max - self.options.min + 1;
+				self._prevValue = null; // self property has to be "null" on start
```
lslachciak, line 118: where _prevValue is used?

TomaszLukawskiSam:
> this works when apopbar is not extended.
> but when appabr is extended double scrolling occurs,
> we may address this in separate issue if you want
> ?

I'll try to fix it before merge this change.

TomaszLukawskiSam:
> this works when apopbar is not extended.
> but when appabr is extended double scrolling occurs,
> we may address this in separate issue if you want
> ?

The issue with AppBar require changes in AppBar widget. We need possibility to lock AppBar expanding during Spin drag. I think separate issue is good idea.

lslachciak (approved):

lslachciak:
I created separate issue https://github.com/Samsung/TAU/issues/1111

Issue #1111, opened 2020-05-20
After extending app bar problem with manipulation of date & time picker 


lmslachciak:
[Schedule App] After extending app bar problem with manipulation of date & time picker 
Using scroller affects also appbar

TomaszLukawskiSam:
Patch has been merged.

lslachciak:
works fine now

PR #1113, opened 2020-05-20
Spin and Appbar: lockExpanding on Spin drag

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1111
[Problem] After extending app bar problem with manipulation of
 date & time picker
[Solution]
 - added new method "lockExpanding()" to Appbar
 - lock Appbar expanding during Spin drag


grzegorz-czajkowski (approved):
Verified. Works fine.

Issue #1108, opened 2020-05-20
[ScheduleApp] Key screen 2 elements overlap


lmslachciak:
Some subtitles on key screen 2 overlap with list content
see "Device off"
![image](https://user-images.githubusercontent.com/2246331/82416442-56225680-9a7a-11ea-9759-e885a97ba223.png)

korneliakobiela:
Please review the issue.

lslachciak:
it is ok now

PR #1114, opened 2020-05-20
Add proper margins to Content Subheader element

korneliakobiela:
[Issue] #1108
[Problem] Some elements overlap
[Solution] - Creating proper paddings (according to guide and App reference)


lslachciak:
I started with testing this PR.
No overlapping occurs but we need to improve spacing according to the guideline
See
image
Vs
image

examples/mobile/ScheduleApp/css/style.css
```diff
@@ -81,4 +81,9 @@
 	position: relative;
 	top: 1px;
 	margin: auto;
+}
+
+.content-subheader {
+	height: 52px;
```
lslachciak: this seems to be a too big gap according to the GUI (2.1)
image
and 
Reference app 
f/[WebApp]Basic_Reference_GUI_V1.0_200311/index.html#artboard9
image

I wander if we should do the styling here - shouldn't it be a TAU library part?

korneliakobiela: Done!

lslachciak (approved):
works fine, looks good

Issue #1132, opened 2020-05-22
Schedule App: add second schedule-list for state after event create


TomaszLukawskiSam:
- add new page: `schedule-list-2.html` (copy from main)
- join `done` button with `schedule-list-2.html`
- remove last list item from `main.html`
- join back button with `main.html`

TomaszLukawskiSam:
Patch has already been merged.

PR #1133, opened 2020-05-22
ScheduleApp: schedule list after event create

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1132
[Problem] Schedule App: add schedule-list for state after event create
[Solution]
 - added new page
 - fix for app flow

### DIFF
--- a/examples/mobile/ScheduleApp/second-list.html
+++ b/examples/mobile/ScheduleApp/second-list.html
@@ -14,10 +14,10 @@
 </head>
 
 <body>
-	<div class="ui-page ui-page-active" id="main">
+	<div class="ui-page ui-page-active" id="second-list">
 		<header>
 			<div class="ui-appbar-left-icons-container">
-				<a href="#" class="ui-btn ui-btn-icon ui-btn-icon-back" data-style="flat" data-rel="back"></a>
+				<a href="#main" class="ui-btn ui-btn-icon ui-btn-icon-back" data-style="flat"></a>
 			</div>
 			<div class="ui-appbar-title-container">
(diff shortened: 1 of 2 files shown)


DohyungLim (commented):
The code is fine in my opinion.

examples/mobile/ScheduleApp/second-list.html
```diff
@@ -0,0 +1,159 @@
...
+	<link href="css/DayOfWeekPicker.css" rel="stylesheet" />
+	<script src="js/DayOfWeekPicker.js"></script>
+	<script src="js/day-indicator.js"></script>
+</head>
+
+<body>
+	<div class="ui-page ui-page-active" id="second-list">
+		<header>
+			<div class="ui-appbar-left-icons-container">
+				<a href="#main" class="ui-btn ui-btn-icon ui-btn-icon-back" data-style="flat"></a>
```
singa2000, line 20: Is it correct the "#main" link  for back key? 
Shouldn't back key go to index.html?

TomaszLukawskiSam: In theory it shouldn't matter because `index.html` has page with id = `main` but flow in router is different. When we know that this page has already exists the page can be changed without load.

singa2000: oh ok. I see.
Thank you for your kind explanation.

lslachciak (approved):
I'm a little skeptical about this PR
We need to remember about removing additional page when saving functionality is done.
If this is needed just  for demo than OK

Issue #1116, opened 2020-05-21
Broken On-Off switch after re-open page


TomaszLukawskiSam:
![image](https://user-images.githubusercontent.com/29534410/82537709-c5677b80-9b4a-11ea-95c7-44037ce6d06f.png)

The issue is cause by rebuild checkbox element inside On-Off Switch.
Can be solved by add `data-role="none"` during build process.

TomaszLukawskiSam:
Patch has already been merged

PR #1135, opened 2020-05-22
OnOffSwitch: fix for broken look after re-open page

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1116
[Problem] Broken On-Off switch after re-open page
[Solution]
 - prevent re-build checkbox created inside on-off switch

[Depends on]
https://github.com/Samsung/TAU/pull/1133


lslachciak (approved):

lslachciak:
works well now

Issue #1143, opened 2020-05-27
[mobile] [ScheduleApp] Add schedule dialog in dark theme


lmslachciak:
[mobile] [ScheduleApp] Add schedule dialog in dark theme looks wrong
![Screenshot from 2020-05-27 10-16-15](https://user-images.githubusercontent.com/2246331/82994942-36071000-a003-11ea-8898-63446abe0f6c.png)

korneliakobiela:
PR merged, please review.

lslachciak:
works well, thank you

PR #1145, opened 2020-05-27
Fix background on dialog element in dark mode

korneliakobiela:
[Issue] #1143
[Problem] Dialog list wasn't visible in dark mode (white background)
[Solution] Define background color dependant from theme


TomaszLukawskiSam (approved):
Looks good to me.

Issue #683, opened 2020-03-02
 [wearable] Presentation Views -> List -> Group Index -> Content Page


lmslachciak:
[wearable] Presentation Views -> List -> Group Index -> Content Page

grzegorz-czajkowski:
@lslachciak, @singa2000 
I have a feeling that in guide example [1], group index with text is presented not in list - there is no:
 * selection (middle element),
 * animation for adjacent element (on list they are usually smaller than middle element)

It looks like this is just ui-content with text elements having separators (group index) with scroll possibility. To ensure this, we will need to add new style, for example, ui-content-group-index.

In the past, we had a possibility to add text on the list by class ui-li-text in example [2] but it does not use TAU Listview since its constructor is excluded in circle-helper.js by checking page id (page-multiline-list).

In my opinion we don't longer support text content on list apart from dedicated sub-text, mulitline presented in guide.
Could you express your opinion on that?

[1] 2.1.1 List: Group index_content_page1_v1.6
[2] examples/wearable/UIComponents/contents/etc/list/list_multiline.html

grzegorz-czajkowski:
Patch merged. Issue can be closed.

PR #1171, opened 2020-06-04
Group Index Content Page (Wearable): add example and style

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/683
[Problem] No support for Group Index on Content Page
[Solution] Share implementation from listview, add example and
	adjust style.


lslachciak (approved):
seems good works good

PR #1174, opened 2020-06-05
Update .gitignore files to ignore release/debug files

lslachciak:
[Issue] N/A
[Problem]
In Android Studio select: Build -> Generate Signed Bundle/APK
debug/release files are visible as new ones in git
[Solution]
Update .gitignore files

### DIFF
--- a/examples/Android/ScheduleApp/.gitignore
+++ b/examples/Android/ScheduleApp/.gitignore
@@ -12,3 +12,5 @@
 /captures
 .externalNativeBuild
 .cxx
+app/release/
+app/debug/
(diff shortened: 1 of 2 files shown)


examples/Android/ScheduleApp/.gitignore
```diff
@@ -12,3 +12,5 @@
 /captures
 .externalNativeBuild
 .cxx
+app/release/
+app/debug/
```
grzegorz-czajkowski, line 16: Is this app dir created in root directory? If yes, it would be better to add slash (/app/.). Do do we need anything from app directory (not sure where apk file is created). If no, suggest adding whole app directory (/app/ or app/).

lslachciak: in app directory there are also sources and build files, we can not ignore them
see: https://github.com/Samsung/TAU/tree/tau_1.2/examples/Android/ScheduleApp/app

grzegorz-czajkowski (approved):
LGTM

Issue #684, opened 2020-03-02
 [wearable] Presentation Views -> List -> Group Index -> left align page


lmslachciak:

grzegorz-czajkowski:
Patch merged. Issue can be closed.

PR #1176, opened 2020-06-05
Group Index Left (Wearable): add example and style

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/684
[Problem] No support for group index left
[Solution] Add implementation and example (content only)

### DIFF
--- a/examples/wearable/UIComponents/contents/presentation-views/list/group-index/multi-left.html
+++ b/examples/wearable/UIComponents/contents/presentation-views/list/group-index/multi-left.html
@@ -1,0 +1,74 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta content="width=device-width, initial-scale=1, user-scalable=no" name="viewport" />
+	<title>
+		Wearable UI
+	</title>
+	<link href="../../../../lib/tau/wearable/theme/default/tau.css" rel="stylesheet" />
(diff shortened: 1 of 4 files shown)


grzegorz-czajkowski:
Implementation for content only. Needs to do the same thing for the list.

DohyungLim (commented):
In my opinion, this is the code without any problems.

grzegorz-czajkowski:
> In my opinion, this is the code without any problems.

Thanks. Soon I will extend this PR to support left adjustment for list as well.

examples/wearable/UIComponents/contents/presentation-views/list/group-index/multi-left.html
```diff
@@ -0,0 +1,74 @@
...
+			</h2>
+		</header>
+		<div class="ui-content">
+			<ul class="ui-listview">
+				<li>
+					<a>
+						List 12
+					</a>
+				</li>
+				<li class="ui-li-group-index-left">
```
TomaszLukawskiSam: I think you can try to do this in another way.
`<div class="ui-li-group-index ui-li-group-index-left">`

grzegorz-czajkowski: Right. Changed examples and simplified styles. Thanks.

TomaszLukawskiSam (approved):
Looks and works very well.

Issue #1161, opened 2020-06-02
Flat button –> ripple effect


lmslachciak:
When Flat button is long-pressed, the shape of the ripple effect is not removed.
see p.9 Web App Review 0.2

PR #1189, opened 2020-06-09
Remove additional effect

korneliakobiela:
[Issue] #1161
[Problem] - When button is clicked longer than usual,
there is additional pulse animation
[Solution] - Remove additional animation

### DIFF
--- a/src/js/core/util/anchorHighlight.js
+++ b/src/js/core/util/anchorHighlight.js
@@ -288,6 +288,7 @@
 			function addButtonInactiveClass() {
 				if (anchorHighlight._buttonTarget) {
 					anchorHighlight._buttonTarget.classList.add(classes.INACTIVE_BTN);
+					anchorHighlight._buttonTarget.classList.remove(classes.ACTIVE_BTN);
 				}
 			}
 


lslachciak:
Is this PR for issue https://github.com/Samsung/TAU/issues/1161 or just part of it?
As mentioned in doc
image
and as you can see in sample app
image

issue is about ripple effect not removed after long press.
After this PR I can still see this issue

src/css/profile/mobile/common/button.less
```diff
@@ -210,8 +210,7 @@ a.ui-btn {
 			&::before {
 				opacity: 1;
 				animation: btn_press_animation_flat_opacity_in linear 100ms,
-					btn_press_animation_flat_opacity_out linear 400ms 100ms,
-					btn_press_animation_flat_scale cubic-bezier(0.33, 0, 0.2, 1) 350ms;
+					btn_press_animation_flat_opacity_out linear 400ms 100ms;
```
lslachciak: this is not needed now

DohyungLim: I agree with @lslachciak .
It is better not to modify it unless it is functionally necessary.

DohyungLim (commented):
1 comment.

lslachciak (commented):
works ok for me, but please double check if this has no side effects for other type of buttons

korneliakobiela:
Checked, I didn't observe any side effects.

lslachciak (approved):

Issue #1191, opened 2020-06-10
[ScheduleApp] Using HW back key on Android causes app quit


lmslachciak:
[ScheduleApp] Using HW back key on Android causes app quit

PR #1192, opened 2020-06-10
Add support for HW back button

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/1191
[Problem] HW back key causes app quit
[Solution] Added HW back key support


TomaszLukawskiSam (approved):
Looks and works very well.

Issue #1168, opened 2020-06-04
Spin / DateTimePicker: stop scroll on swipe - lack of "dragend" event (sometimes)


TomaszLukawskiSam:
The issue cause wrong look of widget:
![obraz](https://user-images.githubusercontent.com/29534410/83730890-003cd980-a64a-11ea-82af-9b6ec98dc6fb.png)

Reproduction path:
- open widget
- try to do fast scroll when finger going out from widget container

lslachciak:
I confirm that issue exists

TomaszLukawskiSam:
I think the issue can be close.

lslachciak:
It is better now, but I still can always reproduce this issue both on desktop as well as on mobile in case of long swipe.
Reproduction:
Start swipe in device On Section
Finish swipe in Device off section below or on the active date
![image](https://user-images.githubusercontent.com/2246331/84652789-7d9b0080-af0c-11ea-86a1-9bb5b6a3dfe0.png)

TomaszLukawskiSam:
@lslachciak 
How you think, can we close the issue?

lslachciak:
yes, now issue can be closed. Thank you for patience

PR #1201, opened 2020-06-16
Spin: fix for stop scroll on swipe

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1168
[Problem] Spin / DateTimePicker stop scroll on long swipe
[Solution]
 - Increasing the number of carousel elements reduces this problem
 - added timeout for lack of dragEnd event and reset drag event detecting

### DIFF
--- a/src/js/core/widget/core/Spin.js
+++ b/src/js/core/widget/core/Spin.js
@@ -57,7 +57,8 @@
 			ROLL_DURATION = 600,
 			DELTA_Y = 100,
 			DRAG_STEP_TO_VALUE = 60,
-			NUMBER_OF_CAROUSEL_ITEMS = 11,
+			NUMBER_OF_CAROUSEL_ITEMS = 13,
+			EVENT_DRAG_END_TIMEOUT = 500,
 
 			/**
 			 * Alias for class Spin
@@ -129,6 +130,7 @@
 				self._lastCurrentIndex = null;
 				self._currentCentralCarouseItem = 0;
 				self._count = 0;
+				self._dragTimeoutHandler = null;
 			},
 
 			WIDGET_CLASS = "ui-spin",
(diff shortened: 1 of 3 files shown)


src/js/core/widget/core/Spin.js
```diff
@@ -57,7 +57,7 @@
 			ROLL_DURATION = 600,
 			DELTA_Y = 100,
 			DRAG_STEP_TO_VALUE = 60,
-			NUMBER_OF_CAROUSEL_ITEMS = 11,
+			NUMBER_OF_CAROUSEL_ITEMS = 13,
```
lslachciak, line 60: is this change still needed?

src/js/core/widget/core/Spin.js
```diff
@@ -677,13 +678,26 @@
...
+				self._dragEnd({
+					detail: {
+						velocityY: e.velocityY,
+						distance: e.distance,
+						direction: e.direction
+					}
+				});
+				self._dragTimeoutHandler = null;
+				ns.event.gesture.Manager.getInstance().resetDetecting();
+			}, 500);
```
lslachciak: how about definig this 500 timeout as const together with other const?
see:
ENABLING_DURATION = 300, // [ms]
			ROLL_DURATION = 600,
			ROLL_DURATION = 600,
			DELTA_Y = 100,
			DELTA_Y = 100,
			DRAG_STEP_TO_VALUE = 60,
			DRAG_STEP_TO_VALUE = 60,
			NUMBER_OF_CAROUSEL_ITEMS = 11,
			NUMBER_OF_CAROUSEL_ITEMS = 13,

lslachciak (commented):
works very well, just two minor comments

src/js/core/widget/core/Spin.js
TomaszLukawskiSam: It is not very necessary but extends touching area for drag/swipe.

src/js/core/widget/core/Spin.js
TomaszLukawskiSam: ok. Done

lslachciak (approved):

Issue #1198, opened 2020-06-15
Toast message - incorrect background


lmslachciak:
see GUI v.2.2 p. 239

lslachciak:
issue fixed

PR #1203, opened 2020-06-16
Fix toast colors

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/1198
[Problem] Wrong background color for toast. Wrong font color for toast
[Solution] Apply proper colors

### DIFF
--- a/src/css/profile/mobile/base.less
+++ b/src/css/profile/mobile/base.less
@@ -77,6 +77,9 @@
 @slider-handler-color: @slider-base-color;
 @slider-level-bar-bg-color: fade(#979797, 30%);
 
+@toast-background-color: fade(#474747, 90%);
+@toast-text-color: @_white;
+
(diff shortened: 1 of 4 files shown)


src/css/profile/mobile/base.less
```diff
@@ -105,6 +108,9 @@
     --slider-value-color: @slider-value-color;
     --slider-handler-color: @slider-handler-color;
 
+    --toast-background: @toast-background-color;
+    --toast-text: @toast-text-color;
```
grzegorz-czajkowski: Since text can have multiple properties (font size etc) I'd prefer the name:
`--toast-text-color`
What do you think?

lslachciak: you are right, fixed

grzegorz-czajkowski (approved):
Looks fine.

Issue #1151, opened 2020-06-01
Arclistview and "more options" does not work


Vascomax:
While adding the "more options" button to a page with an Arclistview, the three dots are shown, but a click on it is not recognized.
I use the default tizen studio 3.7 and it happened in a wearable (tizen 4.0) project.
Is anyone else having this problem too?

HunseopJeong:
@Vascomax,
You need to initialize the selector when opening the page.
You can refer to the UIComponents - contents - moreoptions -moreoptions.js
```
	page.addEventListener("pagebeforeshow", function () {
		var radius = window.innerHeight / 2 * 0.8;

		clickHandlerBound = clickHandler.bind(null);
		handler.addEventListener("click", clickHandlerBound);
		if (tau.support.shape.circle) {
			selector = tau.widget.Selector(elSelector, {itemRadius: radius});
		}
	});
```

Vascomax:
@HunseopJeong,
thanks for your reply. I did combine the UIComponents example for more-options with the TAUlist, so I had that section and moreoptions.js included. I did some tests with changing the list style via circle-helper. 
If I change there ArcListview to SnapListview, without changing anything else in the code, the more options works, as long as the list is not scrolled. The selector opens correctly. But as soon as the list is scrolled, the button disappears and does not come back. (so there seems to be an issue with SnapListview as well...). But I want to use ArcList, so I was focussing on that.
At least with the ArcListview the button is showed all the time, but is not clickable at all.

I added some screenshots:

ArcListviewSituation (button always shown but not working)
![arclistview_more_options](https://user-images.githubusercontent.com/65168100/84213422-e6274f00-aac0-11ea-825b-d1e25a61e82b.png) 
SnapListview before scrolling (button shown and working)
![snaplist_more_before_Scroll](https://user-images.githubusercontent.com/65168100/84213510-238bdc80-aac1-11ea-990c-7ff90d39128e.png)
SnapListview after scrolling (button not shown)
![snaplist_more_after_scroll](https://user-images.githubusercontent.com/65168100/84213770-d3f9e080-aac1-11ea-82ac-41421da198ac.png)

HunseopJeong:
@Vascomax, 
You can fix the issue with z-index.
```
		<div class="ui-content ui-content-padding">
                         // Set the z-index property to the more options button.
			<button class="ui-more" type="button" style="z-index:100;">
				More Options
			</button>
			<ul class="ui-listview">
				<li>
					<a href="#">
[...]

PR #1206, opened 2020-06-16
Change the z-index of more option

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1151
[Problem] More option button does not click when using with arclistview
[Solution] Modify the order of the more option using z-index

### DIFF
--- a/src/css/profile/wearable/changeable/theme-circle/layout.less
+++ b/src/css/profile/wearable/changeable/theme-circle/layout.less
@@ -133,6 +133,7 @@ body {
 		border: 0 none;
 		padding: 0;
 		text-indent: -9999px;
+		z-index: 5;
 		&::before {
 			content: "";
 			position: absolute;


src/css/profile/wearable/changeable/theme-circle/layout.less
```diff
@@ -133,6 +133,7 @@ body {
 		border: 0 none;
 		padding: 0;
 		text-indent: -9999px;
+		z-index: 10;
```
lslachciak: did you consider setting z-index to -1 for ui-listview
to lower priority of it to have more universal solution?

lslachciak (commented):
Looks like CI failing due to tests..

src/css/profile/wearable/changeable/theme-circle/layout.less
HunseopJeong: Is it a more universal solution to lower the priority of the ui-listview?
If it is correct, I will change it.

lslachciak: yes, I was thinking about this as a more universal solution @TomaszLukawskiSam how do you think?

TomaszLukawskiSam: @lslachciak 
I think, @HunseopJeong 's solution is more stable. The listview is a key widget and this modification can have unexpected issues.

lslachciak (approved):

src/css/profile/wearable/changeable/theme-circle/layout.less
```diff
@@ -133,6 +133,7 @@ body {
 		border: 0 none;
 		padding: 0;
 		text-indent: -9999px;
+		z-index: 5;
```
DohyungLim, line 136: Is there any particular reason why the z-index is 5?

DohyungLim (commented):
1 comment.

src/css/profile/wearable/changeable/theme-circle/layout.less
HunseopJeong: I just want to make it more higher than the ui-listview.
But I don't know why it makes the failure for the test.
Need to check the cause.

HunseopJeong:
There was a testing issue.
So I reuploaded the PR(#1255).

Issue #1158, opened 2020-06-02
[UIComponents] Switch has to have constant width, height,  regardless of the switch status 


lmslachciak:
See p. 5-6 Web App Reference v.0.2

TomaszLukawskiSam:
Patch has already been merged.

PR #1209, opened 2020-06-18
(mobile) OnOffSwitch: set constant width and height

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1158
[Problem] Switch has to have constant width, height, regardless of the switch status
[Solution] Style has been changed


HunseopJeong (approved):
LGTM

PR #1216, opened 2020-06-19
Fix the ScheduleApp on/off setting

singa2000:
[Issue] N/A
[Problem]When you select the "Turn device on", it display the menu for "on and off"
[Solution] fix the same behavior with menu


korneliakobiela (approved):
Looks Good! :white_flower:

PR #1218, opened 2020-06-19
Add the Back button function

singa2000:
[Issue] N/A
[Problem] There is not back button in the sample
[Solution] Add the back button in the header


grzegorz-czajkowski (approved):
LGTM.

PR #1219, opened 2020-06-19
Fix small Time Picker on TV

grzegorz-czajkowski:
[Issue] N/A
[Problem] Time Picker is too small when UIComponets app from Mobile
	profile is launched on TV screen (1920x1080).
[Solution] Apply "width: 35%" for screen width greater than 986px.


singa2000 (approved):
It's good working

PR #1220, opened 2020-06-19
Fix ui-content-area for TV

grzegorz-czajkowski:
[Issue] N/A
[Problem] "Repeat weekly" has different width than "Device On/Off" picker.
[Solution] Apply proper width to ui-content-area for Tablet and TV and center
	the element.


korneliakobiela (approved):
LGTM  :sun_with_face:

PR #1221, opened 2020-06-19
(UIComponents) Fix double back action in Dialogs

grzegorz-czajkowski:
[Issue] N/A
[Problem] User is navigated two pages backward adter click "back icon"
	in "Dialog -> General Dialog" or "Dialog -> Picker Dialogs".
[Solution] Exclude ".ui-btn-icon-back" from calling "tau.history.back"
	since it already does it.


korneliakobiela (approved):
Looks well and works wel

HunseopJeong (approved):
LGTM

Issue #1179, opened 2020-06-08
Expose Official NPM Package


AndreasHassing:
Hi TAU team,

It would be really useful to have a (official) NPM package for TAU so we can get updates easier and have a single source of dependencies when using a `package.json` in Tizen projects.

singa2000:
Hi~
Thank you for your suggestion.
We'll check about the npm package and update.

lslachciak:
Hi @AndreasHassing I would like to know more about your use case.
So far we provided examples for front-end web application not based on node.
TAU don't have build system which would allow to pack TAU node module into final front-end web app.

Do you want to use TAU together with other framework/solution which will pack TAU node module into final web app ? E.g. using browserify?

If you want just to use TAU in frontend app without node usage I think better solution would be to provide link to minified TAU.

See jquery usage example:

==========
https://www.npmjs.com/package/jquery

**Browser**
Script tag
<script src="https://code.jquery.com/jquery-3.5.1.min.js"></script>


**Node**
To include jQuery in Node, first install with npm.
npm install jquery

AndreasHassing:
@lslachciak I use webpack in order to use TypeScript and other nice features (also to avoid using Tizen Studio Eclipse). I just place the compiled JavaScript into the Tizen Project and build from there when I need to test on a device.

As an example (not yet using TAU) see this project: https://github.com/AndreasHassing/SmartZensor

If I had a NPM package, I would just add TAU as a dependency.

lslachciak:
@AndreasHassing thank you for sharing use case scenario.
We are analysing effort needed to address this issue.

grzegorz-czajkowski:
TAU has several issue that needs to be solve before we publish it as npm package. The following command fails
`npm install git+https://git@github.com/Samsung/TAU.git#tau_1.2`

- local dependency to console-reported: https://github.com/grzegorz-czajkowski/TAU/commit/a9c1f75a799c1b04a8e0e0ea44234b943b5c6d79
- git hooks can not be installed: https://github.com/grzegorz-czajkowski/TAU/commit/4ff33b6d4da4f3bb42850a08eb43c17d73ec2ef0
 -  'grunt: not found': grunt is defined as devDependencies which means it will be installed by 'npm install' in TAU directory. We need to redefine dependencies vs devDependecies.
 - node 6.16 requirement: to build TAU it's necessary to have node in version 6.16. This force developer to build his own application in this version as well. This is the most problematic issue.

[...]

PR #1235, opened 2020-06-23
Let 'npm install' check node and build TAU

grzegorz-czajkowski:
[Issue] https://github.com/Samsung/TAU/issues/1179
[Problem] 'npm install' does not build TAU
[Solution] Modify preinstall script in order
	to check required node version (check-node.py).
	Build TAU after npm install

### DIFF
--- a/check-node.py
+++ b/check-node.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+
+import json, subprocess, re, sys
+
+current_node_version = []
+expected_node_version = []
+
+def get_version(s):
+	return re.sub('[v\n]', '', s).split('.')
+
+def version_to_str(l):
+	s = '.'
+	s = s.join(l)
+	return s
+
+with open('package.json') as f:
+	data = json.load(f)
+expected_node_version = get_version(str(data['engines']['node']))
+
+try:
+	current_node_version = get_version(subprocess.check_output('node --version', shell=True))
+except subprocess.CalledProcessError as e:
+	print 'Could not get node version, Please install nvm and execute nvm use.'
+	print e.output
+	sys.exit(1)
+
+if len(current_node_version) < 2 or len(expected_node_version) < 2:
+	print 'Not enough information about node version.'
+	sys.exit(1)
+
+# Major and minor version are expected to be equal.
+if current_node_version[0] != expected_node_version[0] or current_node_version[1] != expected_node_version[1]:
+	print 'Current node version is %s but should be %s. Please install nvm and execute nvm use.' % (version_to_str(current_node_version), version_to_str(expected_node_version))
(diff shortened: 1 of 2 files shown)


.nvmrc
```diff
@@ -1 +1 @@
-6.16
+6.16.0
```
lslachciak: do we really need to set "0" ?
when I was adding this file my intention was to add always latest version from 6.16 series

grzegorz-czajkowski: Ok thanks for the comment. I was not aware of this.
Changed approach to compare major and minor versions only.

HunseopJeong (commented):
Is it really necessary to add a python script to check the version?
You can check the node version using the 'https://www.npmjs.com/package/check-node-version'
Is there a reason to use the python?

grzegorz-czajkowski:
> Is it really necessary to add a python script to check the version?
> You can check the node version using the 'https://www.npmjs.com/package/check-node-version'
> Is there a reason to use the python?

We have to check node version in preinstall (before any package is installed) and at this time no package is available. Checking node version in postinstall using node packages does not make sense - we will have to download and install all TAU dependency to check node version.

HunseopJeong:
> > Is it really necessary to add a python script to check the version?
> > You can check the node version using the 'https://www.npmjs.com/package/check-node-version'
> > Is there a reason to use the python?
> 
> We have to check node version in preinstall (before any package is installed) and at this time no package is available. Checking node version in postinstall using node packages does not make sense - we will have to download and install all TAU dependency to check node version.

Okay,, Need to check the node version before installing packages. I see.
Thanks you.

HunseopJeong (approved):
LGTM

check-node.py
```diff
@@ -0,0 +1,32 @@
...
+	print e.output
+	sys.exit(1)
+
+if len(current_node_version) < 2 or len(expected_node_version) < 2:
+	print 'Not enough information about node version'
+	sys.exit(1)
+
+# Major and minor version are expected to by equal.
+if current_node_version[0] != expected_node_version[0] or current_node_version[1] != expected_node_version[1]:
+	print 'Current node version is %s but should be %s. Please install nvm and execute nvm use.' % (current_node_version, expected_node_version)
```
lslachciak: currently message on console is:
_Current node version is ['10', '15', '3'] but should be ['6', '16']._
Can we improve message? e.g. 
_Current node version is 10.15 but should be 6.16_

grzegorz-czajkowski: Fixed, now it prints:
`Current node version is 10.15.3 but should be 6.16. Please install nvm and execute nvm use.`

lslachciak (approved):
finally it works good and seems good too

Issue #1238, opened 2020-06-24
Router: Button with "data-rel='back'" & "href='#'" attributes causes loop on device Tizen 6.0


TomaszLukawskiSam:
Test scenario:
- Install and open UIComponents app
- select 'Dialogs' >  'Picker Dialogs'
- tap back button (left top corner)

Issue:
- The app goes back to previous page but a while goes to the "Picker Dialogs" page

After analize of call stack I found difference between browser on pc and browser on device:
![obraz](https://user-images.githubusercontent.com/29534410/85520240-e8ca8e00-b602-11ea-80ed-e71d4fdb9ae9.png)

TomaszLukawskiSam:
The issue is not 100% repeatable.
In case of proper result of button action, the log looks that:
![obraz](https://user-images.githubusercontent.com/29534410/85532143-bd9a6b80-b60f-11ea-8519-06ecb99f969a.png)

When the issue occurred the log looks different:
![obraz](https://user-images.githubusercontent.com/29534410/85532506-0a7e4200-b610-11ea-8800-968db61c8590.png)

I think the key to the problem is hash change. For example, if we remove `href="#"` from link then the log looks that:
![obraz](https://user-images.githubusercontent.com/29534410/85533288-af008400-b610-11ea-9fe0-5eb434c31888.png)

The `Router.open()` should be called once.

TomaszLukawskiSam:
Hi @lslachciak ,
I made analysis of issue. I'll try to avoid second `Router.open` when there is `data-rel="back"' and hash is changing.

I think that `hash' change should be ignored when we are doing the history back.

TomaszLukawskiSam:
I prepared fix for the issue.
Result in log:
![obraz](https://user-images.githubusercontent.com/29534410/85546766-8632bb80-b61d-11ea-94c0-4c118882269d.png)

TomaszLukawskiSam:
Patch has been merged. I close the issue.

TomaszLukawskiSam:
@lslachciak 
How you think? Move this change to 1.0?

lslachciak:
if we move this change we should also update UIComponents app with @singa2000 commit adding back button
1.0-1.1 series is quite well tested so I don't think we should proapagate this change now

TomaszLukawskiSam:
ok. We can close the issue.

lslachciak:
@singa2000 issue which you reported us was fixed. Please close issue if you confirm on your side too.

[...]

PR #1240, opened 2020-06-24
Router: fix for click on back button - prevent default action

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1238
[Problem] Router: Button with "data-rel='back'" & "href='#'"
  attributes causes loop on device Tizen 6.0
[Solution]
 - TAU had redundant action when click on link with "rel='back'".
  This caused duplicate popstate events and hashchange
  events. In most cases it doesn't matter but sometimes
  these events were in different order and this caused issue
  on device.

### DIFF
--- a/src/js/core/history/manager.js
+++ b/src/js/core/history/manager.js
@@ -105,6 +105,7 @@
 					link = selectorUtils.getClosestBySelector(target, LINK_SELECTOR),
 					href,
 					useDefaultUrlHandling,
+					result = true,
 					options, // this should be empty object but some utils that work on it
 					rel; // require hasOwnProperty :(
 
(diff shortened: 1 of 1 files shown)


src/js/core/history/manager.js
```diff
@@ -130,17 +131,19 @@
 							options.link = link;
 						}
 						history.disableVolatileMode();
+						// mark as handled for back button
```
lslachciak, line 134: code in if's statements seems to be duplicated
can we just write
if (rel === "back" || !triggerStateChange(options))
{
....
}
moreover we will avoid calling the same code twice when both conditions are passed

TomaszLukawskiSam: Yes you are right.

TomaszLukawskiSam: however, this does not work ...
Method `triggerStateChange` has argument `options` with `event`.
This event has to be prevented before call `triggerStateChange` method.

and `triggerStateChange` method has to be called.

lslachciak (approved):

Issue #1244, opened 2020-06-25
Slider: wrong position indicated by slider for "min" != 0 


TomaszLukawskiSam:
![image](https://user-images.githubusercontent.com/29534410/85668699-550ac780-b6bf-11ea-8c2d-b735920c1dd7.png)

When slider option `min` != 0 then slider indicates wrong value

TomaszLukawskiSam:
Patch has already been merged. The issue can be close.

PR #1245, opened 2020-06-25
Slider: fix for min value option

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1244
[Problem] When slider option min != 0 then slider indicates
 wrong value
[Solution]
 Method for set widget value on UI has been changed


HunseopJeong (approved):
LGTM

lslachciak (approved):

PR #1247, opened 2020-06-26
Load only the length of the item

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1236
[Problem] Refer to the index that is more than the length of the data
[Solution] Check the length of the data when loading items


grzegorz-czajkowski (approved):
This fix is for virtual list. LGTM.

lslachciak (approved):

singa2000 (approved):
It looks good.

Issue #1249, opened 2020-06-30
[Wearable] Even after destroy widget, the item remains when using virtual list


HunseopJeong:
Reproduce steps
1. Clone the https://github.com/hykim1/Tizen
2. Click the 'virtual list issue'
3. Back to the home
4. Re-click the 'virtual list issue'

Expected: starts with data from the virtuallist-db-demo-100.js
Actual: starts with data from the virtuallist-db-demo-20.js + from 21 index data from the  virtuallist-db-demo-100.js

PR #1250, opened 2020-06-30
Remove items when destroying the widget

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1249
[Problem] Even though the widget was destroied, items remained
[Solution] Prevent to append items to the element

### DIFF
--- a/src/js/profile/wearable/widget/wearable/ArcListview.js
+++ b/src/js/profile/wearable/widget/wearable/ArcListview.js
@@ -1782,11 +1782,6 @@
 					marqueeWidget = null;
 
 				self._unbindEvents();
-
-				self._items.forEach(function (li) {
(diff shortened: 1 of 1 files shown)


src/js/profile/wearable/widget/wearable/ArcListview.js
```diff
@@ -1782,11 +1782,6 @@
 					marqueeWidget = null;
 
 				self._unbindEvents();
-
-				self._items.forEach(function (li) {
-					self.element.appendChild(li);
-					li.setAttribute("style", "");
-				});
```
HunseopJeong, line 1789: @lslachciak, @TomaszLukawskiSam,
Do you know why it append items to the element here?

Issue #1248, opened 2020-06-29
No sample items in schedule App


lmslachciak:
1. Run Schedule App
2. Observe: No sample items

korneliakobiela:
Development done, please review.

lslachciak:
works fine

PR #1252, opened 2020-06-30
Add default schedule elements.

korneliakobiela:
[Issue] #1248
[Problem] - Default elements don't have type parameter
[Solution] - Add type parameter to defaultStore


lslachciak (approved):
just small nit for commit message
"Defaults elements doesn't have type parameter"
-->
Default elements don't have type parameter

PR #1255, opened 2020-07-01
Change the z-index of more option

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1151
[Problem] More option button does not click when using with arclistview
[Solution] Modify the order of the more option using z-index


DohyungLim (approved):
LGTM.

DohyungLim (commented):
@TomaszLukawskiSam 
I need your opinion on z-index.
If it doesn't matter, merge it.

singa2000 (approved):
It looks good.

PR #1256, opened 2020-07-01
Remove items when destroying the widget

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1249
[Problem] Even though the widget was destroied, items remained
[Solution] Prevent to append items to the element

### DIFF
--- a/src/js/profile/wearable/widget/wearable/ArcListview.js
+++ b/src/js/profile/wearable/widget/wearable/ArcListview.js
@@ -1953,10 +1953,6 @@
 
 				self._unbindEvents();
 
-				self._items.forEach(function (li) {
-					self.element.appendChild(li);
-					li.setAttribute("style", "");
-				});
 				self._items = [];
 
 				// Destroy marquee.


DohyungLim (approved):
LGTM.

src/js/profile/wearable/widget/wearable/ArcListview.js
```diff
@@ -1953,10 +1953,6 @@
 
 				self._unbindEvents();
 
-				self._items.forEach(function (li) {
-					self.element.appendChild(li);
-					li.setAttribute("style", "");
-				});
```
HunseopJeong, line 1959: @TomaszLukawskiSam, @lslachciak ,
Do you know why appending items on the destroy?

TomaszLukawskiSam: When the widget are building the items are moved from element to widget cache (dummy element) and next appended to carousel. On destroy method these elements should be recovered to previous structure.

HunseopJeong: @TomaszLukawskiSam,
I know what you mean.
However, the items added through the virtual list are also restored to the element.
So it makes some issue #1249.

TomaszLukawskiSam: Ok. I see this now. This make sense to remove this code but we need create issue to recovery html structure in future.

TomaszLukawskiSam (approved):
Looks good to me, but please create issue for recovery html structure with link to this change.

HunseopJeong:
@TomaszLukawskiSam,
I created the issue(#1262).

singa2000 (approved):
If it fixed for issue #1249, is it also need to apply in the tau_1.1?
I think he use the tau1.1.
But the modification it looks good.

HunseopJeong:
@singa2000, 
Yes, It would be good to apply to the tau1.1.
But I already tried to upload the PR to the 1.0 branch.
But there is a issue on the CI test.

@lslachciak,
Is there a way to apply PR to other branches such as 1.1, 1.0?

lslachciak:
@HunseopJeong let's see this CI issue.
Please upload patch to 1.0 branch and we will check the CI issue.

Issue #1257, opened 2020-07-02
[Wearable] Scroll end effect appears even if it is not in the last position


HunseopJeong:
TAU version: 1.1.13
Example Source: https://github.com/hykim1/Tizen/tree/master/TAUListTest

Change some code in snaplistview2.js
```
//			listHelper = tau.helper.SnapListStyle.create(list, {animate: "scale", snapListview: false});
//			var snap = listHelper.getSnapList();
//			snap.scrollToPosition(50);
			listHelper = tau.widget.ArcListview(list);
			listHelper.scrollToPosition(50);
```

Reproduce steps
1. Click "scrollToPosition issue"
2. Try to scroll on the scroll animation
3. Scroll to the down side

[Expected]
No end effect until it reaches the end of the page.

[Actual]
End effect appers even if it is not in the last position
![image](https://user-images.githubusercontent.com/29908741/86313794-f3b79b00-bc60-11ea-9c24-0ab2cd28038f.png)

PR #1259, opened 2020-07-02
Initialize the position of the scroller when opening the page

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1257
[Problem] Set the wrong max height of the scroll because of the scroller.scrollTop
[Solution] Initialize the scrollTop when initializing the page


singa2000 (approved):
It looks good to me.

TomaszLukawskiSam (approved):
Verified. I don't see any side effects. Looks good to me.

Issue #1160, opened 2020-06-02
More options list -> position


lmslachciak:
In case of More options list, the popup should place upon the More options button
see p.8 Web App Review 0.2

TomaszLukawskiSam:
Can we close the issue?

lslachciak:
right

PR #1272, opened 2020-07-07
DropDownMenu: Fix for widget postion in AppBar

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1160
[Problem] More option position is wrong when AppBar is extended
[Solution]
 - height of Dropdown Menu placeholder in AppBar has changed to 56px
   because AppBar max height is 56px
 - DropDown Menu placeholder has to be visible during calculation of
   top position
 - Top margin of Menu when AppBar is extended should 0

[Demo]
Normal AppBar
image

Extended AppBar
image

### DIFF
--- a/src/css/profile/mobile/common/dropdownmenu.less
+++ b/src/css/profile/mobile/common/dropdownmenu.less
@@ -95,6 +95,10 @@
 	}
 }
 
+.ui-dropdownmenu-force-display {
+	display: block !important;
+}
+
 .ui-dropdownmenu-native {
 	select {
 		display: block;
@@ -297,4 +301,11 @@
 
 .ui-li-static.ui-li-has-dropdownmenu {
 	height: 60 * @px_base;
+}
+
+.ui-appbar .ui-dropdownmenu-placeholder {
+	line-height: 56 * @px_base;
+}
+.ui-appbar-expanded .ui-dropdownmenu-placeholder {
+	line-height: (56 + 3) * @px_base; // compensate top margin
 }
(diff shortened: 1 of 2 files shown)


src/js/profile/mobile/widget/DropdownMenu.js
```diff
@@ -1000,7 +1000,7 @@
 					elOptionContainer.addEventListener("focusin", self._focusBound); // bubble
 					elOptionContainer.addEventListener("focusout", self._blurBound); // bubble
 					if (ui.screenFilter) {
-						ui.screenFilter.addEventListener("vclick", self._toggleMenuBound);
```
lslachciak, line 1003: was there any particular reason to change vclick to vmousedown?

src/css/profile/mobile/common/dropdownmenu.less
```diff
@@ -95,6 +95,10 @@
 	}
 }
 
+.ui-dropdownmenu-force-display {
+	display: block !important;
```
lslachciak, line 99: I always afraid about using  !important
was there any issue when there was no !important?

src/js/profile/mobile/widget/DropdownMenu.js
TomaszLukawskiSam: Yes, this is intended. The drop down menu has to close immediately after overlay touch. I verified this behaviour in Android app with AppBar (eg. Messages).

src/css/profile/mobile/common/dropdownmenu.less
TomaszLukawskiSam: This is the only way to overwrite inline style. This class in not part of API and it is used only for check position of placeholder.

HunseopJeong (approved):
Looks good to me

lslachciak (approved):

Issue #1278, opened 2020-07-08
Improve Karma test line coverage to 60%


lmslachciak:
Current status as of 2020/07/08
npm run test:karma
53,96%

TomaszLukawskiSam:
Latest status: 
![obraz](https://user-images.githubusercontent.com/29534410/87312435-da480600-c520-11ea-974e-27a05b1ed75e.png)

We need increase test coverage by 780 lines more.

TomaszLukawskiSam:
Subtask: Appbar - increase test coverage (https://github.com/Samsung/TAU/issues/1299)

lslachciak:
Subtask util/path coverage - https://github.com/Samsung/TAU/issues/1305

lslachciak:
Latest status of coverage
Lines:   57.79%

TomaszLukawskiSam:
I think we can close the issue.

lslachciak:
we got it!

PR #1280, opened 2020-07-08
Ensure that karma test coverage will not fall below current coverage

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/1278
[Problem] Test coverage may decrease when adding new code
[Solution] Ensure that Karma fails when lines coverage is dropping below current one.
This should cause CI to fail too.
Current coverage is 53.96


korneliakobiela (approved):
LGTM

Issue #1284, opened 2020-07-08
MoreMenu: wrong top position after content scroll


TomaszLukawskiSam:
![image](https://user-images.githubusercontent.com/29534410/86917775-f5cc9e80-c125-11ea-8621-ce580859467e.png)

TomaszLukawskiSam:
The issue was solved. Thanks.

PR #1287, opened 2020-07-09
MoreMenu: Correct dropdown position after scroll

korneliakobiela:
[Issue] #1284
[Problem] Dropdown position is not correct when content scrolled
[Solution] - Add condition to checking if element isn't overflowed
out the screen


lslachciak (approved):
seems OK, works OK
checked with HommeApp and UIComponents/components/view-controls/dropdownlist/align-to-list.html

Issue #1296, opened 2020-07-10
Add sample of negative TC (nTC) for TAU


TomaszLukawskiSam:
- mark negative test case by `[n]` in name of test case
- test should check attributes in method call
- add the test to Karma runner

TomaszLukawskiSam:
Patch has already been merged.

PR #1297, opened 2020-07-10
Util: added method "tau.throws()" and added tests for it

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1296
    [Problem] Add sample of negative TC for a widget
    [Solution]
     - added new TAU method "throws" which throws TAUException
     - added check if method throws exception in case of
     argument lack
     - added tests for core.throws with negative TC (nTC)
     - added nTC for util.requestAnimationFrame without callback

obraz

obraz

### DIFF
--- a/src/js/core/core.js
+++ b/src/js/core/core.js
@@ -41,7 +41,10 @@
 				args.unshift("[" + rootNamespace + "][" + dateNow.toLocaleString() + "]");
 			},
 			ns = window.ns || window.tau || {},
-			nsConfig = window.nsConfig || window.tauConfig || {};
+			nsConfig = window.nsConfig || window.tauConfig || {},
+			TAUException = function (message) {
+				this.message = message;
+			};
 
 		ns.info = ns.info || {
 			profile: "custom"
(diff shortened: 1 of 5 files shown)


src/js/core/core.js
```diff
@@ -179,6 +186,18 @@
...
 
+		ns._TAUException = TAUException;
+
+		/**
+		 * Method throws TAU exception with given message
+		 * @param {string} message
+		 * @method throws
+		 * @member ns
+		 */
+		ns.throws = function (message) {
```
HunseopJeong, line 197: How about  checking the type of message whether it is string or not?

TomaszLukawskiSam: This is good idea. I'll add the checking.

HunseopJeong (approved):
LGTM

lslachciak (approved):
nit:
you improved lines coverage up to 54 %
I think you could update file
```
tests/karma/all.conf.js
check: {
				global: {
					lines: 54
				}
			}
```
so other PRs will not decrease lines coverage

Issue #1242, opened 2020-06-25
Some urls was set to the wrong location.


HunseopJeong:
Wrong screen is shown when clicking the refresh button on the UIComponents app.
Need to fix the urls of the sub menus.

lslachciak:
@HunseopJeong can we close this issue?

HunseopJeong:
@lslachciak,
Many pages still need to be modified.
I am not enough time to see. 
I'll close it when finished.

PR #1317, opened 2020-07-20
Fix wrong urls of style and script in appbar

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1242
[Problem] Some urls refer to wrong url
[Solution] Modify urls to refer to correct path


TomaszLukawskiSam (approved):
Looks good to me.

singa2000 (approved):
LGTM

Issue #1360, opened 2020-08-11
[HomeApp] [POC] widgets in iFrames


lmslachciak:
Confirm that widgets can be loaded using iframes
Please put exemplary widget content into
webclip subfolder

singa2000:
There is not need the wearable TAU library.
It's enough the mobile library. Would you remove it,please?

lslachciak:
@TomaszLukawskiSam I think this can be closed

TomaszLukawskiSam:
I think the same. We can close the issue.

PR #1362, opened 2020-08-11
(example) HomeApp: change emebeded content of widget to iframe

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1360
[Problem] Confirm that widgets can be loaded using iframes
[Solution]
 - Hotel Breakfast content has been changed to iframe


singa2000:
There is not needed the wearable library.
I think it's enough the mobile library.
Would you remove it, please?

singa2000:
If each web clip have the own tau library, the size is too big.
HomeApp should be light and fast.
I think we need to get the solution for it.
(Maybe developer will release the their web clip with tau library)

lslachciak:
> If each web clip have the own tau library, the size is too big.

Yes, this is one of the biggest drawbacks of the iframe approach. We are aware of it.
In case of few webclips this will mean that TAU library is loaded to memory few times.
We are thinking if there is a better approach

lslachciak:
One of another issues which I see here is size of widgets. They don't fit in app correctly. 
@TomaszLukawskiSam I uderstand this is a problem of iframe which can not set content size?
image

TomaszLukawskiSam:
@lslachciak 
The iframe solution causes a lot of problems: repeatedly embedded TAUs, problems with widget dimensions, problems with graphic themes.

Currently I am preparing a solution based on Custom Elements. There are also problems but different. Mainly a problem with relative url for images https://lists.w3.org/Archives/Public/public-webapps/2014OctDec/0013.html

TomaszLukawskiSam:
This PR is deprecated.

Issue #1361, opened 2020-08-11
[HomeApp] [POC] store installed widgets configuration locally


lmslachciak:
Store locally installed widgets so next start of app will restore their list

lslachciak:
@korneliakobiela can we close issue?

PR #1368, opened 2020-08-13
[Home App] implement storage

korneliakobiela:
[Issue] N/A
[Problem] - Webclips needs storage to store plugins which are installed
[Solution] - Implement store

### DIFF
--- a/examples/mobile/HomeApp/index.html
+++ b/examples/mobile/HomeApp/index.html
@@ -9,7 +9,8 @@
 	<link rel="manifest" href="./manifest.webmanifest">
 	<link rel="shortcut icon" href="images/Icon.png">
 	<script data-build-remove="false" src="lib/tau/mobile/js/tau.js"></script>
-	<script src="js/app.js"></script>
+	<script src="js/clipping-storage.js" type="module"></script>
+	<script src="js/app.js" type="module"></script>
 </head>
 <body>
 	<div class="ui-page ui-page-active" id="main">
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -1,9 +1,13 @@
+import Storage from "./clipping-storage.js";
+
 (function () {
+
 	"use strict";
 	var tau = window.tau,
 
+		storage = new Storage(),
 		webClipList = [],
-		activeWebClipList = [
+		activeWebClipList = storage.readAllFromStorage(Storage.elements.ACTIVEWEBCLIPS) || [
 			"webclip/tv-remote-control",
 			"webclip/restaurant",
 			"webclip/weather"
@@ -37,7 +41,7 @@
 						"webclip/apps-on-tv",
 						"webclip/latest-news",
 						"webclip/now-on-tv"
-					]
+					];
 			})
 			.finally(() => {
 				updateWebClipsUI();
@@ -48,14 +52,16 @@
 	function onPopupSubmit() {
 
 		activeWebClipList = [];
+		storage.refreshStorage(Storage.elements.ACTIVEWEBCLIPS);
 		webClipList.forEach(function (webclip) {
 			const webClipName = getWebClipName(webclip),
 				checkbox = document.querySelector("#" + webClipName);
 
 			if (checkbox.checked) {
-				activeWebClipList.push(webclip)
+				activeWebClipList.push(webclip);
+				storage.writeToStorage(Storage.elements.ACTIVEWEBCLIPS, webclip);
 			}
-		})
+		});
 		updateWebClipsUI();
 		tau.history.back();
 	}
--- a/examples/mobile/HomeApp/js/clipping-storage.js
+++ b/examples/mobile/HomeApp/js/clipping-storage.js
@@ -1,0 +1,24 @@
+class Storage {
+	constructor() {
+		this.activeWebClips = JSON.parse(localStorage.getItem("activeWebClips")) || [];
+	}
+
+	refreshStorage(element, data = []) {
+		this[element] = data;
+		localStorage.setItem(element, JSON.stringify(this[element]));
+	}
+
+	writeToStorage(element, link) {
+		this[element].push(link);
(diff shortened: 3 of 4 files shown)


DohyungLim (approved):
I think no problem.

lslachciak (changes requested):
1. After refreshing button items are not restored
2. Please refer to https://github.com/Samsung/TAU/issues/1361 issue

lslachciak:
due to last week changes and PR: https://github.com/Samsung/TAU/pull/1373 we need to store/restore list of active weblips on app start: activeWebClipList

examples/mobile/HomeApp/index.html
```diff
@@ -46,6 +47,250 @@
 			-->
 			<!-- Home App Plugins (end) -->
 
+<<<<<<< HEAD
```
singa2000: Would you remove the unnecessary code, please?

examples/mobile/HomeApp/js/app.js
```diff
@@ -77,6 +81,7 @@
 	function onPopupSubmit() {
 		getWidgetsState();
 		toggleWidgetsState();
+		storage.readFromUI(event.target)
```
lslachciak: readFromUI returns value. Is it used anywhere?

lslachciak (changes requested):
As I wrote in previous comment we need to store/restore activeWebClip list not UI elements values

examples/mobile/HomeApp/js/app.js
korneliakobiela: Done!

examples/mobile/HomeApp/js/clipping-storage.js
```diff
@@ -0,0 +1,25 @@
+class Storage {
+	constructor() {
+		this.webclips = JSON.parse(localStorage.getItem("webclips")) || [];
```
lslachciak: we shouldn't store webclips list in local storage. this is dynamic list which is changed frequently

examples/mobile/HomeApp/js/clipping-storage.js
```diff
@@ -0,0 +1,25 @@
...
+		this.activeWebClips = JSON.parse(localStorage.getItem("activeWebClips")) || [];
+	}
+
+	refreshStorage(element, data) {
+		this[element] = data;
+		localStorage.setItem(element, JSON.stringify(this[element]));
+	}
+
+	writeToStorage(element, link) {
+		this[element].push(link);
```
lslachciak, line 12: you should replace table not expand it
as a result you add additional elements to table

lslachciak (changes requested):
need to make changes in order to realise scenario, after refresh of webappitems are not stored correctly

examples/mobile/HomeApp/js/clipping-storage.js
korneliakobiela: I expand it only locally but in the next line rewrite to local storage.

lslachciak: each time onPopupSubmit() will be called elements will be appended to table this[element]
This will make this[element] containing the same elements many times.

lslachciak:
Please rebase this PR, current version outputs many errors on console. I'm not able to customize widgets to check this issue.
There are no errors like this in latest version on github

`Refused to apply style from 'http://106.120.63.110:3000/webclip/css/style.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
`

```
10app.js:81 Uncaught TypeError: Cannot read property 'checked' of null
    at app.js:81
    at Array.forEach (<anonymous>)
    at HTMLButtonElement.onPopupSubmit (app.js:77)

```

lslachciak (changes requested):
still after refreshing webpage active webclip configuration is not restored

1. Open HomeApp
2. Customize widget
3. Click refresh button

korneliakobiela:
Ok, now it should work.

examples/mobile/HomeApp/js/app.js
```diff
@@ -14,7 +18,7 @@
 	}
 
 	function loadWebClipList() {
-		const requestURL = "api/webcliplist";
+		const requestURL = "http://localhost:3000/api/webcliplist";
```
lslachciak: please don't change this url to localhost.
app can be run in other context.

examples/mobile/HomeApp/js/clipping-storage.js
```diff
@@ -0,0 +1,24 @@
+class Storage {
+	constructor() {
+		this.activeWebClips = JSON.parse(localStorage.getItem("activeWebClips")) || [];
+	}
+
+	refreshStorage(element, data = []) {
```
lslachciak, line 6: refreshStorage is never called with second param, what is it for?
why call it?
as I checked in debugger this functions always set empty this[element]

lslachciak (changes requested):

examples/mobile/HomeApp/js/clipping-storage.js
korneliakobiela: For now, this is used only for clear storage (assign empty array to it), but please consider, that changes are everything in this project and this method allows to rewrite whole storage with new data. It can be quite useful. But if you want, I can remove this feature

singa2000:
I just have a question about the storage.
If the device home is closed or device is turn off, the storage data is keeping or disappear?

Would you also consider the web clip ordering rule, please?

lslachciak (approved):

lslachciak:
> I just have a question about the storage.
> If the device home is closed or device is turn off, the storage data is keeping or disappear?

Storage is on the mobile device not TV, so it will exist as long as phone is not wiped out
"the stored data is saved across browser sessions"
Turning off Home device (TV) will not have influence

https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage
> 
> Would you also consider the web clip ordering rule, please?

I'll create separate issue for this

PR #1371, opened 2020-08-18
 Read WebClip list 

lslachciak:
[Issue] #1366
[Problem] No example to read webclip list
[Solution] Provide exemplary code using fetch API

### DIFF
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -6,7 +6,10 @@
 				news: false,
 				nowontv: false
 			}
-		};
+		},
+		// TODO: remove below directive after webClipList use
+		// eslint-disable-next-line no-unused-vars
+		webClipList = [];
 
 	function getWidgetsState() {
 		var latestNews = document.querySelector("#latest-news-check"),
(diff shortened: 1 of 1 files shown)


TomaszLukawskiSam (approved):
Patch works good but we will issue with CORS when app and app server work on different port. This issue can be solved by another patch.

examples/mobile/HomeApp/js/app.js
```diff
@@ -49,6 +52,16 @@
 		loadWeatherJS();
 	}
 
+	function loadWebClipList() {
+		const requestURL = "api/webcliplist";
```
DohyungLim, line 56: as Tomasz said, it contains CORS issues.
If the URL is run locally, you need to consider the issue.

DohyungLim (commented):
1 comment.

examples/mobile/HomeApp/js/app.js
lslachciak: ok, I understand case although this is specific use case.

app is served always with API on the same port so by default no issue occurs. This is only in case when  API will be served separately from app, but this is not our use case I think

TomaszLukawskiSam: ok

Issue #1372, opened 2020-08-21
[HomeApp] Add WebClips to the app based on data from JSON


TomaszLukawskiSam:

lslachciak:
I think we can close this issue

PR #1374, opened 2020-08-25
Allow customization of widget list

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/1372
[Problem] No possibility to customize widget list using add window
[Solution] Update activeWebClipList according to user choice

Signed-off-by: Lukasz Slachciak <l.slachciak@samsung.com>

PR #1375, opened 2020-08-27
Synchronize list of active web clips with available web clips 

lslachciak:
[Issue] #1372
[Problem] When webclip removed from server it is still available to user
[Solution] Update active web clips list according to the available webclips

### DIFF
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -2,18 +2,11 @@
 	"use strict";
 	var tau = window.tau,
 
-		webClipList = [
-			"webclip/tv-remote-control/",
-			"webclip/restaurant/",
-			"webclip/weather/",
-			"webclip/apps-on-tv/",
-			"webclip/latest-news/",
-			"webclip/now-on-tv"
-		],
+		webClipList = [],
 		activeWebClipList = [
-			"webclip/tv-remote-control/",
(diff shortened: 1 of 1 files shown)


lslachciak:
should be merged after https://github.com/Samsung/TAU/pull/1374

examples/mobile/HomeApp/js/app.js
```diff
@@ -48,6 +41,28 @@
...
+				activeWebClipList.forEach(function (webclip) {
+					if (!webClipList.includes(webclip)) {
+						activeWebClipList.splice(activeWebClipList.indexOf(webclip), 1);
+					}
+				})
+			})
+			.catch(() => {
+				// use default webclip list is sth wrong
+				webClipList =
+					[
```
singa2000, line 54: I have a question.
If  web clip is removed from the server, is it also updated in the "webClipList"? So "webClipList" don't have the a removed web clip ?

lslachciak: both webClipList and activeWebClip list will not have removed web clip.
So it will not be presented in web app.
But please note that this list is updated only on web app start, so after removing webclip form server web app need to be restarted.

We can think about additional notification mechanism from server if we want to update webclip list when app is already launched

lslachciak:
PR rebased

korneliakobiela (approved):
LGTM

Issue #1377, opened 2020-08-28
Multiple pages in single html, issue with multiple selectors and rotarydetent


niteraven7:
TAU version 1.0.27
Platform: wearable 4.0

I'm building a multiple page web application using a single html file. If I move between the pages and load up various selectors in popups (I don't believe popups have anything to do with it) multiple rotarydetent listeners are being created on document. Therefor when I move the bezel it moves all of the other selectors. From what I understand I should be destroying the selectors as I move between the pages or popups. I tried this... the issue is when I destroy the selector and I try to recreate selector the rotarydetent does not work for that selector any longer. I have spent days trying to solve this issue and there doesn't seem to be a way to solve it.

Any way to handle this?

Thank you,
Kyle

niteraven7:
Maybe I can temporarily enable/disable the Selector instead of destroying?

niteraven7:
The enable/disable idea proved to be correct and is working. But I'll leave this open since I think there may be a bug with destroying and recreating Selectors with rotarydetent listeners

TomaszLukawskiSam:
Hi,
I was able to reproduce this issue. There are two issues. The first is bug in destroy method (as @niteraven7 wrote) because the widget's initial structure is not restored. The second issue is the hidden selector is receiving events eg. rotary detent. Is also possible that some properties are common for different instances of selector.
I'll try to solve the issue by disable event listeners on hidden selector.

TomaszLukawskiSam:
Hi @niteraven7,

I prepared fix for TAU solving the issue. This is not big change but works. You can test it, I attached zip file with `tau.js`.

niteraven7:
Hi @TomaszLukawskiSam! Thanks so much for quickly addressing and identifying the issue(s). The TUA framework is truly impressive and I'm discovering new features every day. Creating a working and functional UI on a 1" screen is truly a fun challenge! :)

Have you fixed both issues? The multiple Selector background listening does need to be addressed by either the framework or developer. If the create and destroy and recreate Selector bug can be fixed then I'll at least have the control!

ps. I couldn't find the link the the tau.js file but do see your commits.

TomaszLukawskiSam:
Hi @niteraven7 ,
The fix only applies to ignoring events on hidden selector.
The file with TAU has been attached to PR (https://github.com/Samsung/TAU/files/5149023/tau-wearable.zip)

I created new issue (https://github.com/Samsung/TAU/issues/1383) for destroy method because is much more complicated to do.

PR #1378, opened 2020-08-31
(wearable) Selector: ignore events when widget is hidden

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1377
[Problem] Events dedicated to active instance of Selector has impact on
 other instances of Selector even when they are hidden.
[Solution]
 - fix: hidden Selector instance ignores events

[tau-wearable.zip](https://github.com/Samsung/TAU/files/5149023/tau-wearable.zip)

### DIFF
--- a/src/js/profile/wearable/widget/wearable/Selector.js
+++ b/src/js/profile/wearable/widget/wearable/Selector.js
@@ -811,43 +811,46 @@
 			prototype.handleEvent = function (event) {
 				var self = this;
 
-				switch (event.type) {
-					case "dragstart":
-						self._onDragstart(event);
-						break;
-					case "drag":
-						self._onDrag(event);
-						break;
-					case "dragend":
-						self._onDragend(event);
-						break;
-					case "click":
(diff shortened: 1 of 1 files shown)


src/js/profile/wearable/widget/wearable/Selector.js
```diff
@@ -811,43 +811,46 @@
...
-						self._onTouchEnd(event);
-						break;
-					case "tizenhwkey":
-						event.stopPropagation();
-						self._onHWKey(event);
-						break;
-					case "swipe":
-						self._onSwipe(event);
-						break;
+				// Events on the hidden selector are ignored
```
HunseopJeong: I think 'Events on the hidden selector should be ignored' is more proper.

TomaszLukawskiSam: ok, I'll change this.

HunseopJeong (approved):
LGTM

Issue #1384, opened 2020-09-02
[HomeApp] API returns files


lmslachciak:
API: /api/webcliplist returns both files and directories
It should return directories only

lslachciak:
PR done

PR #1385, opened 2020-09-02
Omit files when listing webclip directory content

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/1384
[Problem] Server API for webclip list returns both files and directories
[Solution] Use fs.lstatSync(path_string).isDirectory() in order to distinguish
directories as suggested in [1]

[1] - https://stackoverflow.com/questions/15630770/node-js-check-if-path-is-file-or-directory#15630832


TomaszLukawskiSam (approved):
Verified. Looks good to me.

Issue #1356, opened 2020-08-03
Chips: bottom margin should be 0


TomaszLukawskiSam:
Spacing on the bottom of chips.
According to the guideline bottom space should be 0 but this assumption is difficult to reach.

![image](https://user-images.githubusercontent.com/29534410/89016116-73945c00-d318-11ea-87c2-4547a9ba06ff.png)

_Originally posted by @TomaszLukawskiSam in https://github.com/Samsung/TAU/issues/1330#issuecomment-667003120_

TomaszLukawskiSam:
Patch has already been merged. Issue can be close.

PR #1413, opened 2020-09-15
Chips: fix bottom space, wrong icon color and size

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1356
[Problem] Chips: bottom margin should be 0
[Solution]
 - fix for wrong icon image
 - fix for bottom space

obraz

### DIFF
--- a/src/js/profile/mobile/widget/Chip.js
+++ b/src/js/profile/mobile/widget/Chip.js
@@ -94,6 +94,8 @@
 				 * @static
 				 */
 				eventUtil = ns.event,
+				COLOR_CLASS_PREFIX = "ui-color-",
+				COLOR_CLASS_REGEX = /ui-color-[^\s]+/g,
 
 				Chip = function () {
 					var self = this;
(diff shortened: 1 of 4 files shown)


src/js/profile/mobile/widget/Chip.js
```diff
@@ -265,8 +259,9 @@
 				ui.button = ui.button || element.querySelector(selectors.button);
 
 				ns.widget.Button(ui.button, {
-					icon: self._getButtonIcon()
+					icon: self.options.icon
 				});
+				this._setButtonColor(ui.button);
```
HunseopJeong: _setButtonColor seems that doesn't have a parameter.
Is it correct?

TomaszLukawskiSam: Thanks for review. This is mistake and method name also make confusion. I'll change this.

korneliakobiela (approved):
Looks good, works good

Issue #1426, opened 2020-09-21
Drawer: right space is too big on TV


TomaszLukawskiSam:
![obraz](https://user-images.githubusercontent.com/29534410/93742024-b32b4600-fbed-11ea-95d7-a951c8049e48.png)

TomaszLukawskiSam:
Patch has been merged. I close the issue.

PR #1427, opened 2020-09-21
Drawer: change widget width on TV resolution

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1426
[Problem] Drawer: right space is too big on TV
[Solution]
 - widget width for big screen is fixed to 360px

[Screenshot]
obraz


singa2000 (approved):
It looks good.
Thanks

PR #1434, opened 2020-09-23
Correctly remove popup list items

lslachciak:
[Issue] N/A
[Problem] We shouldn't enumerate items and remove them at the same time. This can lead to unexpected behaviour

[Solution] Remove items in modern way. See more at [1]

[1] https://stackoverflow.com/questions/3955229/remove-all-child-elements-of-a-dom-node-in-javascript#3955238


korneliakobiela (approved):
Looks good to me

Issue #1430, opened 2020-09-23
[HomeApp] WebClip label should be based on description metainfo


lmslachciak:
[HomeApp] WebClip label (on add webclip dialog) should be based on description metainfo

korneliakobiela:
Done! Please review

lslachciak:
works fine. thank you

PR #1435, opened 2020-09-23
Set Webcliplist labels as manifest descriptions

korneliakobiela:
[Issue] #1430
[Problem] webcliplist labels doesn't look natural
[Solution] Fetch webclips manifests

### DIFF
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -174,14 +174,22 @@ import Storage from "./clipping-storage.js";
 				input.setAttribute("type", "checkbox");
 				input.setAttribute("id", webClipName);
 
-				label.setAttribute("for", "latest-news-ckeck");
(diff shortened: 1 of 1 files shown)


examples/mobile/HomeApp/js/app.js
```diff
@@ -174,14 +174,22 @@ import Storage from "./clipping-storage.js";
 				input.setAttribute("type", "checkbox");
 				input.setAttribute("id", webClipName);
 
-				label.setAttribute("for", "latest-news-ckeck");
+				label.setAttribute("for", webClipName);
```
lslachciak, line 177: do we need to set this attribute for?

korneliakobiela: This is the accessibility enhancement. User can tap/click on the label and then set/unset the option. Without this user have to hit the button, which can be small.

lslachciak (approved):

Issue #1437, opened 2020-09-24
[HomeApp] Apply ordering rules to app


lmslachciak:
Apply following ordering rules to HomeApp:

1. WebClips for active apps are always before WebClips for inactive
2. Control cards are always before service cards

lslachciak:
Currently we are discussing need of control card, so please implement point 
`1. WebClips for active apps are always before WebClips for inactiv`
first

lslachciak:
Ordering by app active was done in app
    // check apps order
        appsList = appsList.sort(function (app1, app2) {
            return (app1.isActive) ?
                (app2.isActive) ? 0 : -1 : 1
        });

lslachciak:
Basing ordering by active app is already done

PR #1441, opened 2020-09-29
Order webclips in every app by type

korneliakobiela:
[Issue] #1437
[Problem] Control cards should be before service cards in every app
[Solution] Sort cards in apps

### DIFF
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -122,6 +122,15 @@ import Storage from "./clipping-storage.js";
 				(app2.isActive) ? 0 : -1 : 1
 		});
 
+		updateManifests();
+
+		appsList = appsList.forEach((app) => {
+
+			app.webClipsList.sort((a, b) => {
+				return a.manifest.cardType - b.manifest.cardType;
+			});
+		})
+
 		// order has been changed
 		if (currentOrder !== appsList.reduce(function (prev, app) {
 			return prev + app.appID;
@@ -148,6 +157,18 @@ import Storage from "./clipping-storage.js";
 			tau.log("nothing change");
 		}
 
+	}
+
+	function updateManifests() {
+		for (const app of appsList) {
+			for (const webclip of app.webClipsList) {
+				fetch(`${webclip.url}/manifest.json`)
+					.then((data) => data.json())
+					.then((manifest) => {
+						webclip.manifest = manifest;
+					});
+			}
+		}
 	}
 
 	function addWSListener(wsPort) {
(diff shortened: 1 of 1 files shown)


examples/mobile/HomeApp/js/app.js
```diff
@@ -125,6 +125,24 @@ import Storage from "./clipping-storage.js";
 				(app2.isActive) ? 0 : -1 : 1
 		});
 
+		// check webclips order in app
+		appsList = appsList.forEach((app) => {
+			fetch(`${app.webClipsList[0].url}/manifest.json`)
```
TomaszLukawskiSam: How often the manifest will change? Is this possible during the app runtime? I think will be better to read the manifest if not exists in apps list yet and store in apps list for next use. This reduce number of reading.

examples/mobile/HomeApp/js/app.js
```diff
@@ -125,6 +125,24 @@ import Storage from "./clipping-storage.js";
...
+				.then((data) => data.json())
+				.then((manifest) => {
+					let temp;
+
+					if (manifest.cardType == "service") {
```
TomaszLukawskiSam: If you had previously stored the data in an array, now you could use it in `.sort()` method for array

TomaszLukawskiSam (changes requested):
I can see the optimization of this code.

examples/mobile/HomeApp/js/app.js
```diff
@@ -125,6 +126,24 @@ import Storage from "./clipping-storage.js";
...
+			if (app.isActive) {
+				const appWebClips = new Map();
+
+				app.webClipsList.forEach((webclip) => {
+					appWebClips.set(webclip.url, manifestsList[webclip.url]);
+				});
+
+				Promise.all(appWebClips).then((value) => {
+					app.webClipsList.sort((a, b) => {
+						return (value[a.url] > value[b.url]) ? b : a;
```
TomaszLukawskiSam: I don't understand reason of  sort by `url` property. I thought we want sort by card type "service" | "control".

examples/mobile/HomeApp/js/app.js
```diff
@@ -125,6 +126,24 @@ import Storage from "./clipping-storage.js";
 				(app2.isActive) ? 0 : -1 : 1
 		});
 
+		updateManifests();
+
+		appsList = appsList.forEach((app) => {
+			if (app.isActive) {
```
lslachciak: only sorting for active apps?
we should sort
"Control cards are always before service cards" regardless if app is active or not.

.eslintrc.js
```diff
@@ -340,5 +340,6 @@ module.exports = {
 	},
 	"parserOptions": {
 		"sourceType": "module",
+		"ecmaVersion": 8
```
lslachciak: I think this change is already on repo. Please rebase.

examples/mobile/HomeApp/js/app.js
```diff
@@ -153,6 +170,15 @@ import Storage from "./clipping-storage.js";
 
 	}
 
+	function updateManifests() {
+		for (const app of appsList) {
+			for (const webclip of app.webClipsList) {
+				manifestsList[webclip.url] = fetch(`${webclip.url}/manifest.json`)
```
lslachciak: can we just keep result of json fetch in appsList object rather than on separate manifestList?
We already store additional properties there as isSelected for webclip.
This would be stored/restored from local storage together with apps list and could make 
updateAppsList simpler

TomaszLukawskiSam:
This PR is outdated and will no longer be needed.

Issue #1442, opened 2020-10-02
npm install fail


lmslachciak:
Reproduction steps

```
#clean repo
git clean -dfx
#install dependencies
npm install
```
```
Observe:

Loading "svg-sprites.js" tasks...ERROR
>> TypeError: Object.values is not a function
Warning: Task "svg-sprites" not found. Use --force to continue.

```

korneliakobiela:
I've not found why the issue is in particular. This is a problem with some dependencies of dependencies (dependency hell :imp: ). I suppose that is a bad version of CSS-WHAT and that caused failing.

HunseopJeong:
I got a same error. How about upversioning the node?
We used the v6.16. I think It is too old version.

lslachciak:
We already did some tries regarding node upverisoning but this is quite big change, not sure if this is right for 1.2 release.
Also this would make sense if this would solve this issue.

egorshulga:
I suppose that's because `Object.values()` is available only starting with Node v7 ([see](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values#Browser_compatibility)). v6.16.0 is way too old.

Anyway, I was lucky enough to build the library with Node v7

korneliakobiela:
Yeah, I see it. I am working on make it works well again. Now I see for version Node 8.17 works well, but our tests started to fail.

lslachciak:
I think this issue was solved with https://github.com/Samsung/TAU/pull/1444 so we can close it.
For bumping node I opened new issue: https://github.com/Samsung/TAU/issues/1445

PR #1444, opened 2020-10-06
Fix the issue that 'npm install' fail

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1442
[Problem] Node.js v6.16 doesn't support Object.values()
[Solution] Add the harmony option to enable the Object.values()

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,19 +4,19 @@
   "description": "Tizen Advanced UI",
   "version": "1.2.7",
   "scripts": {
-    "build": "grunt build",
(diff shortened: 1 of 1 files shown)


HunseopJeong:
This PR is a hotfix to install the package correctly.
I think we need to consider the upversion of node.

singa2000 (approved):
I'm confirmed. It's good working.
Thanks

TomaszLukawskiSam (approved):
Looks good to me.

lslachciak:
We need this for further development so I'm merging PR

PR #1446, opened 2020-10-06
add basic ui

SwimingKim:


HunseopJeong:
@SwimingKim, Thanks for uploading the PR.
However, in my opinion, in the case of a Basic app, it refers to a basic app that has only title content for each profile.
In addition, when uploading a PR, it would be good to write an description of the content that led to this PR.

Issue #1456, opened 2020-10-07
[HomeApp] isSelected property does not have effect


lmslachciak:
If "**isSelected**" property is set on server it doesn't have effect locally

Repro:
1. Clear appStroage
2. modify server
diff --git a/examples/mobile/HomeAppServer/app.js b/examples/mobile/HomeAppServer/app.js
index 1e9265c2..bd82521d 100644
--- a/examples/mobile/HomeAppServer/app.js
+++ b/examples/mobile/HomeAppServer/app.js
@@ -17,7 +17,8 @@ var logger = require("morgan"),
                "isActive": true,
                "webClipsList": [
                        {
-                               url: "webclip/apps-on-tv"
+                               url: "webclip/apps-on-tv",
+                               isSelected: "true"
                        },
3. run HomeApp
expected:
apps-on-tv webclip should be selected

PR #1457, opened 2020-10-07
Properly set isSelected attributed. Update apps list

lslachciak:
[Issue] https://github.com/Samsung/TAU/issues/1456
[Problem] isSelected attributed is not received from server
[Solution] 2 changes made:
1. Change updateapplists mechanizm - wrong assignement to undefined value was done
2. Use default list only when failed to receive data from server, so attribute isSelected will not be skipped
Also updated default active values to see changing order of clips.

### DIFF
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -16,7 +16,7 @@ import Storage from "./clipping-storage.js";
 			"webClipsList": [
 				{
 					url: "webclip/apps-on-tv",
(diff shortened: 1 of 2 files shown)


TomaszLukawskiSam (approved):
Looks good to me.

lslachciak:
Additionally we discussed with @TomaszLukawskiSam that we should make one improvement
https://github.com/Samsung/TAU/issues/1458
but this will be a separate PR

PR #1462, opened 2020-10-12
Change the example to fit the content

HunseopJeong:
[Issue] N/A
[Problem] The name of "Circularly horizontal dot" and the example
          content do not match
[Solution] Modify the example to match with the name

### DIFF
--- a/examples/wearable/UIComponents/contents/navigationelements/pageindicator/pageindicator-circular.js
+++ b/examples/wearable/UIComponents/contents/navigationelements/pageindicator/pageindicator-circular.js
@@ -1,0 +1,48 @@
+/*global tau */
+(function () {
+
+	var page = document.getElementById("pageIndicatorCircularPage"),
+		changer = document.getElementById("hsectionchanger"),
+		sections = document.querySelectorAll("section"),
+		sectionChanger,
+		elPageIndicator = document.getElementById("pageIndicator"),
+		pageIndicator,
+		pageIndicatorHandler;
+
+	/**
+	 * pagebeforeshow event handler
+	 * Do preparatory works and adds event listeners
+	 */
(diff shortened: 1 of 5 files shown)


examples/wearable/UIComponents/contents/navigationelements/pageindicator/pageindicator-circular.js
```diff
@@ -19,6 +19,7 @@
 		pageIndicator.setActive(0);
 		// make SectionChanger object
 		sectionChanger = new tau.widget.SectionChanger(changer, {
+      circular: true,
```
TomaszLukawskiSam: wrong indentation

TomaszLukawskiSam: This property indicates that changing of section is in loop... I have on mind the after last item section changer goes to first section. Is this intended change?

HunseopJeong: Yes, it is.
There was no difference between the two existing examples(pageindicator-horizontal & pageindicator-circle).
So I change from the pageindicator-circle to the pageindicator-circular.
Do you know the difference between the two previous examples?

TomaszLukawskiSam: Difference is in behaviour. For `circular=false`, when you change section to the last section you cannot change more. You can change only back. Otherwise, it loops to the first.

HunseopJeong: That's what I intended.
Anyway, I modified it to fit the content.

TomaszLukawskiSam (approved):
Looks good to me.

Issue #1468, opened 2020-10-15
[HomeApp] Apps on TV is being removed


lmslachciak:
Reproduction steps:

1. Run http-server
2. visit http://X.Y.Z.Q:8080/examples/mobile/HomeApp/
3. clean applicaiton Local storage (F12 + Application --> local Storage)
4. refresh page
5. click + icon
6. click OK
Observe:
Apps on TV webclip is no longer visible

TomaszLukawskiSam:
We can close this issue.

PR #1469, opened 2020-10-15
HomeTVApp: fix for checkbox ids in popup

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1468
[Problem] Apps on TV is being removed
[Solution]
 - checkbox ids have to be more unique


lslachciak (approved):
I can confirm, works fine now

PR #1475, opened 2020-10-21
HomeApp: update appslist from diff

TomaszLukawskiSam:

### DIFF
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -75,9 +75,9 @@ import Storage from "./clipping-storage.js";
 
 			fetch(requestURL)
 				.then((response) => response.json())
-				.then((data) => {
-					addWSListener(data.wsPort);
-					resolve(data.apps);
+				.then((response) => {
+					addWSListener(response.data.wsPort);
+					resolve(response.data.apps);
 				})
 				.catch((e) => {
 					reject(e);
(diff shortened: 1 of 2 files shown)


examples/mobile/HomeAppServer/app.js
```diff
@@ -96,11 +133,11 @@ function runPing(ws) {
 		// @test
 		// active app rotation
 		if (pingCount % 3) {
-			apps[0].isActive = false;
-			apps[1].isActive = true;
+			sampleDiff[0].isActive = false;
+			sampleDiff[1].isActive = true;
```
lslachciak, line 137: how about rotating action field so we can verify remove too
e.g
if (ping...) {
sampleDiff[0].action = "add";
sampleDiff[1].action = "remove";
}
else {
sampleDiff[0].action = "remove";
sampleDiff[1].action = "add";
}

examples/mobile/HomeApp/js/app.js
```diff
@@ -131,11 +131,77 @@ import Storage from "./clipping-storage.js";
...
+				delete remoteApp.action;
+
+				if (!localApp) { // add new
+					homeApp.appsList.push(remoteApp);
+				} else { // update local app
+					/**
+					 * @todo
+					 * Which properties we need update
+					 */
+					localApp.isActive = remoteApp.isActive;
```
lslachciak, line 162: I see two problems here
1. when app is already on list but server sends new set of webclips, they will not be updated
2. when app is already on list but server sends isSelected property (e.g. in case of ads) - it will not be updated

examples/mobile/HomeApp/js/app.js
```diff
@@ -321,11 +374,13 @@ import Storage from "./clipping-storage.js";
...
 					if (webClip.isSelected) {
-						homeApp.addControlCard({
-							title: webClip.manifest.description,
-							href: "#nowa-karta",
-							icon: "images/Icon.png"}
-						);
+						if (!document.querySelector("[data-title='" + webClip.manifest.description + "']")) {
+							homeApp.addControlCard({
+								title: webClip.manifest.description,
+								href: "#nowa-karta",
```
lslachciak: "nowa-karta" can we change this to English?

lslachciak (commented):
please check comments in review

examples/mobile/HomeAppServer/app.js
TomaszLukawskiSam: I think this is good idea but should be realized in other issue.

examples/mobile/HomeApp/js/app.js
TomaszLukawskiSam: I'll change this.

examples/mobile/HomeApp/js/app.js
TomaszLukawskiSam: 1.This method assumes that server sends diff not whole list.
2. We can create new issue.

examples/mobile/HomeApp/js/app.js
TomaszLukawskiSam: Fixed.

lslachciak (approved):

Issue #1454, opened 2020-10-07
HomeAppTV: rebuild method `updateWebClipsUI` so that the list is updated instead of being cleared and built


TomaszLukawskiSam:

lslachciak:
@TomaszLukawskiSam I think you did it as a part of https://github.com/Samsung/TAU/pull/1467

TomaszLukawskiSam:
No, it is still to do. We can move this issue to Milestone 8

TomaszLukawskiSam:
The issue has already been fixed.

PR #1477, opened 2020-10-22
HomeApp: rebuild webClips UI without clear and build

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1454
[Problem] method `updateWebClipsUI` cause blinking
[Solution]
     - method has been rebuild to update items
    
### DIFF
--- a/examples/mobile/HomeApp/js/app.js
+++ b/examples/mobile/HomeApp/js/app.js
@@ -310,7 +310,7 @@ import Storage from "./clipping-storage.js";
 	function onPopupSubmit() {
 		homeApp.appsList.forEach(function (app) {
 			app.webClipsList.forEach(function (webClip) {
-				const webClipName = getWebClipName(webClip.url),
+				const webClipName = getWebClipNameFromUrl(webClip.url),
 					checkbox = document.getElementById("popup-checkbox-" + webClipName);
 
 				webClip.isSelected = checkbox.checked;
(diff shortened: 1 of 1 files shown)


examples/mobile/HomeApp/js/app.js
```diff
@@ -392,9 +430,7 @@ import Storage from "./clipping-storage.js";
 		tau.engine.createWidgets(webclipsContainer);
 	}
 
-	//TODO: provide mechanism for getting web clip name from webClip meta data
-	//		and separate from getting ID
-	function getWebClipName(webClip) {
+	function getWebClipNameFromUrl(webClip) {
```
lslachciak, line 436: I think we should leave comment about getting webclipName from metadata and spearate it from getting webclip ID
we can have ID duplication problems in future

examples/mobile/HomeApp/js/app.js
```diff
@@ -349,17 +349,38 @@ import Storage from "./clipping-storage.js";
 	}
 
 	function updateWebClipsUI() {
-		var current = document.querySelectorAll(".ui-card"),
-			webclipsContainer = document.getElementById("web-clips");
+		var webclipsContainer = document.getElementById("web-clips"),
+			currentWebClipsCards = [].slice.call(webclipsContainer.querySelectorAll(".ui-card[data-url],.ui-card[data-src]")),
```
lslachciak, line 354: [].slice.call is not obvious,
please add some comment about this code here.

lslachciak (commented):
I tested PR with HomeAppServer .
It works good. no blinking when active app is changed.
Some minor comments in code

examples/mobile/HomeApp/js/app.js
TomaszLukawskiSam: ok

examples/mobile/HomeApp/js/app.js
TomaszLukawskiSam: Done

lslachciak (approved):

examples/mobile/HomeApp/js/app.js
```diff
@@ -369,7 +391,24 @@ import Storage from "./clipping-storage.js";
 			});
 		});
 
-		// activate / deactivate mini control cards
+		// set proper order of cards
+		// @todo change inline styles to css class after merge HomeApp branches
+		webclipsContainer.style.display = "flex";
+		webclipsContainer.style.flexDirection = "column";
+		currentWebClipsCards = [].slice.call(webclipsContainer.querySelectorAll(".ui-card[data-url],.ui-card[data-src]"));
```
singa2000, line 398: Just I have a question.
I know about the "data-src" of ui-card. It's for card resource path like "webclip/apps-on-tv".
What's for use "data-url"?

TomaszLukawskiSam: Hi,
Cards may be in two states one before load of card content (`data-src`) and after load of content (`data-url`). We need check both attributes.

singa2000: ok. I see. Thank you for your kind explanation.

singa2000 (approved):

PR #1480, opened 2020-10-29
Card : Update the Container/Favorite card according to GUI Guide

singa2000:

### DIFF
--- a/examples/mobile/UIComponents/webclip/link-container/webclip.html
+++ b/examples/mobile/UIComponents/webclip/link-container/webclip.html
@@ -9,22 +9,22 @@
 	</div>
 	<div class="ui-content" data-scroll="x">
 		<div class="ui-container">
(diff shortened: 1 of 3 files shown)


TomaszLukawskiSam (commented):
The PR looks good but commit message is incorrect. 
You should add commit message according to  https://github.com/Samsung/TAU/blob/tau_1.2/CONTRIBUTING.md

lslachciak (approved):
This PR is OK, but one more fix is needed for ui-card container.
Ttile has currently 18px in browser. It should be 16px This is because other style is applied over this defined in 
src/css/profile/mobile/common/card.less 

image
image

singa2000:
@lslachciak 
Thanks for your commnets. I fixed what you said.
inspector
title

lslachciak (approved):

Issue #1495, opened 2020-11-05
HomeApp: general control - basic remote control  scroll issue


singa2000:
After https://github.com/Samsung/TAU/pull/1487, basic remote control is good working.
But when it scroll up/down or left/right, the screen is moving.

There is touch-pad in the Basic Control, so touch-pad need the event "touchmove", "touchstart" "touchend" event.
When scrolling up and down, the screen height is exactly fit, so there should not be any movement.
And when scrolling left/right, can you distinguish it from the tap switching event?

singa2000:
It's good working.
We can close the issue.
Thanks.

singa2000:
The PR is merged so I can close it

PR #1496, opened 2020-11-05
HomeApp: stop scrolling on interaction with touchpad area

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1495
[Problem] when it scroll up/down or left/right, the screen is moving.
[Solution]
 - developer has to make own touch event handler for touchpad area


HunseopJeong (approved):
LGTM

Issue #1497, opened 2020-11-05
Fix "Cannot read property 'element' of undefined"


HunseopJeong:
[How to reproduce]

1. Open "UIComponents" of wearable profile
2. Assist Views -> More options
3. Check the console window

Uncaught TypeError: Cannot read property 'element' of undefined
    at Listview.prototype._selectItem (tau.js:37657)
    at Listview.prototype.handleEvent (tau.js:38052)
    at Object.trigger (tau.js:1344)
    at Page.prototype.trigger (tau.js:8372)
    at Page.prototype.onShow (tau.js:11530)
    at deferredFunction (tau.js:8691)
    at Object.deferred.resolve (tau.js:8819)
    at HTMLDivElement.oneEvent (tau.js:8829)

HunseopJeong:
PR was merged.

PR #1498, opened 2020-11-05
Fix the "Uncaught TypeError" on the UIComponent

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1497
[Problem] Uncaught TypeError occurred
[Solution] Changed to select item only when item exists


TomaszLukawskiSam (approved):
Looks good to me.

Issue #1500, opened 2020-11-05
Fix the "Cannot read property 'carouselElement' of undefined"


HunseopJeong:
[How to reproduce]
1. Open "UIComponents" of wearable profile
2. Assist Views -> More options
3. Drag the page
4. Check the console window

tau.js:37023 Uncaught TypeError: Cannot read property 'carouselElement' of undefined
    at Listview.prototype._carouselUpdate (tau.js:37023)
    at Listview.prototype._onTouchStart (tau.js:37456)
    at Listview.prototype.handleEvent (tau.js:38072)

HunseopJeong:
PR was merged.

PR #1501, opened 2020-11-05
Fix the "Uncaught TypeError" on UIComponent

HunseopJeong:
[Issue] https://github.com/Samsung/TAU/issues/1500
[Problem] Cannot read property 'carouselElement' of undefined
[Solution] Update only when the number of carousel items is valid


singa2000 (approved):
It looks good

PR #1503, opened 2020-11-06
 HomeApp : Disable the expanding app bar

singa2000:
[Problem] Expanding app bar is enabled
[Solution] Set "data-expanding-enabled=false"
[Issue] N/A


TomaszLukawskiSam (approved):
Looks good to me.

Issue #1512, opened 2020-11-25
HomeApp: Basic Control - modify touchpad GUI


TomaszLukawskiSam:
- borders area with arrows allows trigger left / right / top / down events
- Touch pad area:
   - swipe up/down/left/right on touch area triggers move the cursor up / down / left / right
   - double tap: triggers OK

TomaszLukawskiSam:
We can close the issue.

PR #1513, opened 2020-11-25
HomeApp: Basic control - touchpad update

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1512
[Problem] Basic Control - touchpad needs GUI update
[Solution]
 - buttons on borders has been updated

[Screenshot]
obraz

### DIFF
--- a/examples/mobile/HomeApp/css/app-remote-control-basic.css
+++ b/examples/mobile/HomeApp/css/app-remote-control-basic.css
@@ -108,6 +108,16 @@
     width: 100%;
     height: 210px;
     justify-items: center;
+    background-color: #ebebeb;
+    border-radius: 32px;
+    margin-top: 10px;
+}
+
+.app-remote-control .app-4way-direction .ui-btn.ui-btn-flat.ui-btn-icon {
+    max-width: 100%;
+    width: 100%;
+    height: 100%;
+    min-width: 32px;
 }
 
 .app-remote-control-icon {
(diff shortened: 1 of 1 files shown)


examples/mobile/HomeApp/css/app-remote-control-basic.css
```diff
@@ -108,6 +108,16 @@
     width: 100%;
     height: 210px;
     justify-items: center;
+    background-color: #ebebeb;
+    border-radius: 32px;
+    margin-top: 10px;
+}
+
+.app-remote-control .app-4way-direction .ui-btn.ui-btn-flat.ui-btn-icon {
+    max-width: 100%;
```
lslachciak, line 117: do we need to set both width and max-width at the same time with the same units?
I understand that when width and max-width refer to different units.
e.g. 
width: 50px
max-width: 100%
it makes sense.

see also
https://stackoverflow.com/questions/6456468/css-width-and-max-width
```
if width > max-width use max-width
if max-width > width use width
```

TomaszLukawskiSam: Your comment make sense but I have to overwrite styles from other selectors. These styles are already set in different places, eg. `max-width` is already set to `32px`, `width: 58px`, `min-width: 58px`. These buttons are very specific because are placed on borders of touchpad and I need expand their sizes to touchpad dimensions.

lslachciak (approved):

Issue #1508, opened 2020-11-25
Homa App: Create TV Plus Channels as service card


TomaszLukawskiSam:

TomaszLukawskiSam:
Patch has already been merged. I close the issue.

PR #1514, opened 2020-11-25
HomeApp: new webclip TV Plus Channels

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1508
[Problem] add TV Plus Channels webclip
[Solution]
  - new webclip has been added

[Screenshot]
obraz


singa2000:
Would you also add the container, please?
container

examples/mobile/HomeApp/webclip/tv-plus-channels/css/style.css
```diff
@@ -0,0 +1,4 @@
+.app-icon {
```
singa2000: Should we  define it as css separately?
Is not include the Card elements?

TomaszLukawskiSam: > Would you also add the container, please?
> container

I'll change this.

TomaszLukawskiSam:
> Would you also add the container, please?
> container

I'll change this.

examples/mobile/HomeApp/webclip/tv-plus-channels/css/style.css
TomaszLukawskiSam: We have to define somewhere the icon image `../images/title/tvplus.png`.
You don't want this separate file. Do you prefer url to icony to be defined in webclip.html?

singa2000: How about your think?
I think it's better to be defined in webclip.html by .
In case of list icon, we also add in the html.

					<div class="ui-li-icon">
						
					</div>

TomaszLukawskiSam: It is ok for me. I'll change this.

singa2000 (approved):
LGTM

Issue #1515, opened 2020-11-26
Home App : General Control


singa2000:
General Control  is different with Guide

TomaszLukawskiSam:
Can we close the issue?

singa2000:
Sure. I'll close it.

PR #1516, opened 2020-11-26
HomeApp: General Control update

singa2000:
[Issue]https://github.com/Samsung/TAU/issues/1515
[Problem] General Control is different with GUI Guide
[Solution] Update the appbar container


TomaszLukawskiSam (approved):
Looks good to me.

Issue #1493, opened 2020-11-04
SubTab: Scroll items from outside screen if take too much space 


TomaszLukawskiSam:
![obraz](https://user-images.githubusercontent.com/29534410/98093596-6abba500-1e88-11eb-9be2-8ae6d6a8df49.png)

TomaszLukawskiSam:
Patch has already been merged.

PR #1524, opened 2020-11-30
SubTab: scroll sub tabs if take too much space

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1493
[Problem] Scroll items from outside screen if take too much space
[Solution]
 - added new feature

### DIFF
--- a/src/css/profile/mobile/common/subtab.less
+++ b/src/css/profile/mobile/common/subtab.less
@@ -92,9 +92,25 @@
 	}
 
 	&:not(.ui-sub-tab-static) {
+		overflow-x: scroll;
+		margin-left: 0;
+		margin-right: 0;
+		width: 100%;
+
+		&::-webkit-scrollbar {
+			display: none;
+		}
+
 		ul {
 			li {
-				min-width: 105 * @px_base;
+				min-width: 93 * @px_base;
+
+				&:first-child {
+					margin-left: 24 * @px_base;
+				}
+				&:last-child {
+					margin-right: 24 * @px_base;
+				}
 			}
 		}
 		&.ui-sub-tab-landscape {
--- a/examples/mobile/HomeApp/control.html
+++ b/examples/mobile/HomeApp/control.html
@@ -36,7 +36,7 @@
 		</header>
 
 		<div class="ui-content ui-tabs">
-			<div class="ui-sub-tab ui-sub-tab-static">
+			<div class="ui-sub-tab">
 				<ul>
 					<li class="ui-li-anchor">
 						<a class="ui-tab-active" href="#">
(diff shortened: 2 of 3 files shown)


src/css/profile/mobile/common/subtab.less
```diff
@@ -92,9 +92,25 @@
 	}
 
 	&:not(.ui-sub-tab-static) {
+		overflow-x: scroll;
+		margin-left: 0;
+		margin-right: 0;
+		width: 100%;
+
+		&::-webkit-scrollbar {
+			display: none;
```
DohyungLim, line 101: Is there a case of scroll?
Is there a reason for none between display and visibility?

src/css/profile/mobile/common/subtab.less
```diff
@@ -92,9 +92,25 @@
...
+		}
+
 		ul {
 			li {
-				min-width: 105 * @px_base;
+				min-width: 93 * @px_base;
+
+				&:first-child {
+					margin-left: 24 * @px_base
```
DohyungLim: add semicolon at last.

src/css/profile/mobile/common/subtab.less
```diff
@@ -92,9 +92,25 @@
...
+				}
+				&:last-child {
+					margin-right: 24 * @px_base
```
DohyungLim: ditto.

DohyungLim (commented):
3 comment.

src/css/profile/mobile/common/subtab.less
TomaszLukawskiSam: On reference movie scrollbar does not display. The `visibility` will reserve place for scrollbar I think it is not necessary in this case. Do you think `visibility` will be better?

src/css/profile/mobile/common/subtab.less
TomaszLukawskiSam: Done.

src/css/profile/mobile/common/subtab.less
TomaszLukawskiSam: Done.

src/css/profile/mobile/common/subtab.less
TomaszLukawskiSam: > Is there a case of scroll?

You mean if I really need a `overflow-x: scroll`. It seemed right to me. Otherwise, you will have to click on the next items to get to the last items.

singa2000: I also agree with Tomasz. In that case scrollbar does not display.

singa2000 (approved):
LGTM

Issue #1499, opened 2020-11-05
Card: conflict of Card header and Page appbar


TomaszLukawskiSam:
Header in Card is expanding with Page Appbar.

![obraz](https://user-images.githubusercontent.com/29534410/98230545-93609f00-1f5b-11eb-9967-d83567521dd4.png)

TomaszLukawskiSam:
Patch has already been merged.

PR #1526, opened 2020-12-01
AppBar: change widget selector to fix conflict with card header

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1499
[Problem] Card: conflict of Card header and Page appbar
[Solution]
 - sector detailing for appbar solves the problem

[Screenshot]
obraz


singa2000 (approved):
LGTM

PR #1553, opened 2020-12-14
Fix the Uncaught TypeError

HunseopJeong:
[Issue] N/A
[Problem] Uncaught Error Occurred
[Solution] Remove undefined _unbindEvents

### DIFF
--- a/src/js/core/widget/core/Card.js
+++ b/src/js/core/widget/core/Card.js
@@ -172,16 +172,6 @@
 
 			};
 
-			/**
-			 * Destroy Card component
-			 * @method _destroy
-			 * @member ns.widget.core.Card
-			 * @protected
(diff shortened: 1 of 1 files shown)


HunseopJeong:
image

src/js/core/widget/core/Card.js
```diff
@@ -179,7 +179,6 @@
 			 * @protected
 			 */
 			prototype._destroy = function () {
```
TomaszLukawskiSam: I think you can remove whole method if this method does nothing.

HunseopJeong: Done

TomaszLukawskiSam (approved):
Looks good to me.

Issue #1562, opened 2021-01-04
Card: Subheader - disable word wrap


TomaszLukawskiSam:
![obraz](https://user-images.githubusercontent.com/29534410/103540371-e2fb1280-4e99-11eb-9aa6-d4eba62db77f.png)

PR #1563, opened 2021-01-04
Card: Disable word wrap in subheader

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1562
[Problem] Text in card header is wrapped
[Solution]
 - text wrap has been disabled

[Screenshot]
obraz

### DIFF
--- a/src/css/profile/mobile/common/card.less
+++ b/src/css/profile/mobile/common/card.less
@@ -20,6 +20,7 @@
 		.ui-content-subheader {
 			display: flex;
 			align-items: center;
+			white-space: nowrap;
 
 			&::after {
 				content: "";


src/css/profile/mobile/common/card.less
```diff
@@ -20,6 +20,7 @@
 		.ui-content-subheader {
 			display: flex;
 			align-items: center;
+			white-space: nowrap;
```
lslachciak, line 23: shouldn't we have 
 text-overflow: ellipsis in such case?

TomaszLukawskiSam: This is more complicated. We cannot use `text-overflow` without `overflow: hidden` but `overflow: hidden` stay in opposite to `display: flex` and doesn't work. So, we should wrap text in `span` element and define `max-width` for its. In this place would be nice use marquee widget but it all is out of scope guideline.

lslachciak: If this change is according to the guidelines I think this is OK, I was just afraid that text will be cut unexpectedly :)
But someone should approve this change too. I'm here just open source reviewer :)

lslachciak (approved):

src/css/profile/mobile/common/card.less
TomaszLukawskiSam: Thank you for review :)

lbardeli (approved):
LGTM

singa2000 (approved):
LGTM

Issue #1595, opened 2021-02-12
The focus on the listview moves every 2 items


TomaszLukawskiSam:

TomaszLukawskiSam:
PR has been merged. I close the issue.

PR #1597, opened 2021-02-15
KeyboardSupport: fix for focus change calculation

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1595
[Problem] The focus on the listview moves every 2 items
[Solution] Change of condition for filter of focusable elements
     which very near by selves.
    

singa2000 (approved):
fix for TV keyboard support

Issue #1596, opened 2021-02-15
Fix the change of focus by remote control for new components


TomaszLukawskiSam:
- switch on-off
- master switch on-off
- main tabs
- sub tabs

TomaszLukawskiSam:
PR has been merged. I close the issue.

PR #1598, opened 2021-02-16
Styles of focused state for components

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1596
[Problem] Fix the change of focus by remote control for
    new components
[Solution]
   - added style for: on-off switch, radio, sub tabs

### DIFF
--- a/src/css/profile/mobile/base.less
+++ b/src/css/profile/mobile/base.less
@@ -129,4 +129,6 @@
     --on-off-switch-on-disabled-track-background: @on-off-switch-on-disabled-track-background;
     --on-off-switch-off-track-background: @on-off-switch-off-track-background;
     --on-off-switch-off-disabled-track-border: @on-off-switch-off-disabled-track-border;
+
+    --focus-outline-color: rgb(69, 143, 255);
 }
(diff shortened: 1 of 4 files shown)


src/css/profile/mobile/base.less
```diff
@@ -129,4 +129,6 @@
     --on-off-switch-on-disabled-track-background: @on-off-switch-on-disabled-track-background;
     --on-off-switch-off-track-background: @on-off-switch-off-track-background;
     --on-off-switch-off-disabled-track-border: @on-off-switch-off-disabled-track-border;
+
```
singa2000, line 132: Is it correct the profile/mobile profile?
If so, in case of mobile touch , it don't display the focus outline?

TomaszLukawskiSam: The focus on mobile profile is disabled.

Focus outline will only work in tv profile, this is enabled in /src/js/profile/tv/config.js
```
	ns.setConfig("keyboardSupport", true);
```

HunseopJeong (approved):
LGTM

Issue #1634, opened 2021-03-16
Wrong text position in button based on anchor element


TomaszLukawskiSam:
Text should be in the middle of button.

![obraz](https://user-images.githubusercontent.com/29534410/111270751-89009100-8630-11eb-84a9-7c33532ecfb0.png)

TomaszLukawskiSam:
Patch has already been merged.

PR #1635, opened 2021-03-16
Button: fix for text align of button on anchor element

TomaszLukawskiSam:
[Issue] https://github.com/Samsung/TAU/issues/1634
[Problem] Text should be in the middle of button
[Solution]
 - added new css styles for button


singa2000 (approved):
For text align

PR #1678, opened 2021-07-19
Upload new images for mde companion app

HunseopJeong:


singa2000 (approved):
for companion app guide

singa2000:
I think it also need to update the latest video.wgt

PR #1692, opened 2021-10-08
GridView: Fit image to grid item without crop

TomaszLukawskiSam:
New css class has been added to Gridview widget
The `ui-gridview-image-icon` fits image size to the grid item container.

The GridView widget detects icons and adds `ui-gridview-image-icon` to image.

obraz

### DIFF
--- a/src/css/profile/mobile/common/gridview.less
+++ b/src/css/profile/mobile/common/gridview.less
@@ -164,6 +164,18 @@ tau-gridview {
 		}
 	}
 
+	&.ui-gridview-image-icon .ui-gridview-item img,
+	.ui-gridview-item.ui-gridview-image-icon img {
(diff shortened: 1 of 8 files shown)


src/css/profile/mobile/common/gridview.less
```diff
@@ -164,6 +164,18 @@ tau-gridview {
 		}
 	}
 
+	&.ui-gridview-image-icon .ui-gridview-item img,
```
singa2000: If the girdview image is mixed with general image and icon 
and I don't know which image will be on the grid , how about it?
I think in that case I can't use the "ui-gridview-image-icon".

singa2000 (approved):
Fix the image in the gridview

